PR #91, opened 2022-03-14
feat: add-learner-access

Jamal-majid-ti:
add-learner-access ui component


packages/learner-access/src/components/LoadUserLearning.tsx
```diff
@@ -0,0 +1,272 @@
...
+  GlobalTypes,
+  hydrateContent,
+  HydratedContentItem,
+  formatTime
+} from '@thoughtindustries/content';
+import { ArchiveButton } from './MutationCallingButtons';
+import { Tooltip } from '../Assets/Tooltips';
+import { useTranslation } from 'react-i18next';
+import useLearnerAccess from '../use-context';
+import { t } from 'i18next';
```
rjschill87, line 29: Don't need to import `t` here since we're using the `useTranslation` hook on line 42. You can just update line 42 to 


```javascript
const { i18n, t } = useTranslation();
```

packages/learner-access/src/components/LoadUserLearning.tsx
```diff
@@ -0,0 +1,272 @@
...
+            </div>
+
+            <div className="col-start-11 col-span-2 text-right">
+              <button
+                onClick={() => {
+                  window.location.href = window.location.hostname + '/learn/course/' + item.slug;
+                }}
+                className="bg-active-blue text-white rounded-sm cursor-pointer inline-block font-normal text-xs m-0 py-[0.15rem] px-4 relative text-center no-underline ease-in-out border-active-blue font-sans transition duration-200 leading-5"
+              >
+                {item.availabilityStatus == 'completed'
```
rjschill87, line 240: @Jamal-majid-ti My bad not noticing this earlier -- since you're passing an item already handled by `hydrateContent`, you can just display `{item.callToAction` since all this logic is already handled there:

https://github.com/thoughtindustries/helium/blob/2f6f495301bd7c4b596a19b1d5a5112a97ce4a56/packages/content/src/utilities/hydrate-content/hydrate-content.ts#L101

packages/learner-access/src/components/LoadUserLearning.tsx
```diff
@@ -237,15 +236,15 @@ const LoadUserLearning = ({ query, kind, sort }: LoadedComponentProps): JSX.Elem
                 }}
                 className="bg-active-blue text-white rounded-sm cursor-pointer inline-block font-normal text-xs m-0 py-[0.15rem] px-4 relative text-center no-underline ease-in-out border-active-blue font-sans transition duration-200 leading-5"
               >
-                {item.availabilityStatus == 'completed'
+                {item.callToAction == 'completed'
```
rjschill87, line 239: @Jamal-majid-ti Apologies for the comment clarity; I meant the entire block could be replaced with `{item.callToAction}`. There doesn't need to be any status checks, as `getCallToAction` will handle status checks, content types, etc., and return the appropriate translation:

https://github.com/thoughtindustries/helium/blob/2f6f495301bd7c4b596a19b1d5a5112a97ce4a56/packages/content/src/utilities/hydrate-content/hydrate-content.ts#L154

packages/learner-access/src/components/LoadUserLearning.tsx
```diff
@@ -175,7 +174,7 @@ const LoadUserLearning = ({ query, kind, sort }: LoadedComponentProps): JSX.Elem
           </div>
         </div>
         <div className=" archive mt-2 row-end-4 text-black-light relative">
-          {item.availabilityStatus !== 'completed' && (
+          {item.callToAction !== 'completed' && (
```
rjschill87, line 177: I don't believe the Archive button is gated by the content status, so think we can just remove this entirely.

https://github.com/thoughtindustries/ti/blob/24bacf0e12d1df4c405a296ad370a9db8b770ca7/assets/javascripts/lms/templates/components/ti-dashboard-access-expandable-item.hbs#L158

rjschill87:
Re-evaluating. Closing for now as the branch has gone stale.

PR #141, opened 2022-07-26
Build homepage catalogpage

AlexONeillTI:


tooling/template-base/components/ProductInformationCards.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import logo from '../renderer/trees1.png';
+
+const LearningInfo = (props: {
+  heading: string;
+  subheading: string;
+  leftParagraphHeading: string;
```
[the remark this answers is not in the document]
AlexONeillTI: solution pushed

tooling/template-base/components/Navigation/MobileMenu.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <div className="">
+      <ul className="space-y-2">
+        <div className="flex py-5">
+          <div className="h-11 w-11">
+            <img src={avatar} />
+          </div>
+          <div className="mx-3 my-auto">
+            <button className="hover:text-blue-700">
+              <a href="/sign-in">My Profile</a>
```
rjschill87, line 14: Think it'd be weird for to direct users clicking "My Profile" to a sign in page. If they're already signed in, we should direct them to `/learn/profile`. Which brings me to another point - with this and Line 29 in this component, there's an assumption that the user is signed in, which may not be the case. You'll want to pass the `CurrentUser` down through props or use `usePageContext` to access the current user, and if there is one then conditionally render this and the "Sign Out" button. Would recommend checking out and implementing in the other NavBars as well.

tooling/template-base/components/Navigation/CatalogMobileNavBar.tsx
```diff
@@ -0,0 +1,95 @@
...
+                    <img src={xicon} />
+                  </button>
+                ) : (
+                  <button
+                    id="icon"
+                    type="submit"
+                    className="block h-9 w-9 md:hidden"
+                    onClick={() => setNavbar(!navbar)}
+                  >
+                    <img src={avatar} />
```
rjschill87, line 40: Users can upload their own profile pics, so it doesn't quite work to use a hardcoded image. If you pass the `CurrentUser` down or fetch it from context, it should have an `asset` property (provided the user has uploaded a picture). That's the asset we'll want to use in place of the hardcoded avatar image. 

In the event they don't have an image uploaded, we've defaulted to using the user's initials

https://github.com/thoughtindustries/ti/blob/5e738eddc579a0406e8a6fc1b4d759e48c17e96e/assets/javascripts/lms/templates/components/ti-avatar.hbs#L25

https://github.com/thoughtindustries/ti/blob/5e738eddc579a0406e8a6fc1b4d759e48c17e96e/assets/javascripts/lms/components/ti-avatar.js#L12

rjschill87 (commented):
K @AlexONeillTI Think this is my last round. Other than the comments left above, these should be the last couple things:

- [x] Some `.DS_Store` files in root and `/tooling`; can delete those.
- [x] Let's bring all the CTA component variants under the `/CTA` directory.
- [ ] Some of those one-off components like `Hero` and `ButtonLink`, think we can either put them in their own respective directory (like you have for `Footer`).
- [ ] Looks like the `LearnerAccess` component and `signin.page.tsx` got pulled into this PR, can remove those.
- [ ] Checkout the `locales/translations.json` file from staging
- [ ] Delete the `tailwind.css~Stashed changes` file.
- [ ] Looks like there's duplicates of png and svg files. We don't really need both. We can omit the `Overlay.svg`, but let's try and use SVGs for the dropdown, icons, etc. Should be examples in the other packages of using SVGs, but happy to get you pointed if needed.

tooling/template-base/components/Navigation/MobileMenu.tsx
AlexONeillTI: what link would I use for the help page?

AlexONeillTI: "and if there is one then conditionally render **this** and the "Sign Out" button."
what is "this" referring to?

tooling/template-base/components/Navigation/CatalogMobileNavBar.tsx
AlexONeillTI: is the `asset` property part of the `currentUser`? I can't seem to see it in `types.ts`

tooling/template-base/components/Navigation/NavBar.tsx
```diff
@@ -0,0 +1,32 @@
...
+import CurrentUserSmallScreenNavBar from './CurrentUserSmallScreenNavBar';
+import CurrentUserNavBar from './CurrentUserNavBar';
+import UserLoginNavBar from './UserLoginNavBar';
+
+const NavBar = () => {
+  const pageContext = usePageContext();
+  const { currentUser } = pageContext;
+
+  let navbar;
+  console.log(currentUser?.asset);
```
rjschill87, line 12: :bomb:

packages/catalog/src/types.ts
```diff
@@ -22,6 +22,8 @@ export interface CatalogResultsProps {
   companyDefaultLocale?: string;
   /** currency code to format price */
   currencyCode?: string;
+  /** number of items to return */
+  numberOfContentItems?: number;
```
rjschill87, line 26: Since this prop only exists in your custom component and not in the original catalog package, you'll want to extend the `CatalogResultsProps` (or the `CatalogProps`) interface in your custom component.

tooling/template-base/components/Navigation/UserLoginNavBar.tsx
```diff
@@ -14,7 +14,7 @@ export default function UserLoginNavBar() {
           <div className="flex items-center justify-between md:block">
             {/* logo */}
             <div className="">
-              <Logo />
+              <Logo logoLink="" style="" />
```
rjschill87, line 17: Not sure where the additional props came from? Don't believe the `Logo` component makes any use of those.

AlexONeillTI: Hey, the Logo component is actually supposed to have props as the styling may change such as size based on use case. But I think I should remove the linkUrl as it will always redirect to homepage unless you can think of another use case?

AlexONeillTI: that's a good approach, will implement now

tooling/template-base/components/CatalogAndAggreg/Aggregations.tsx
```diff
@@ -0,0 +1,123 @@
+import React, { ReactNode, useState } from 'react';
+import { useCatalog } from '@thoughtindustries/catalog';
+import clsx from 'clsx';
```
rjschill87, line 3: Can you add this to `template-base`'s `package.json`? It works now because the workspace dependencies are linked, but would have issues when installing the `template-base` and running it on its own. Can install by running `npm i clsx -w tooling/template-base` from the `helium` root.

AlexONeillTI: hey Rob, just pushed there

rjschill87 (approved):

PR #143, opened 2022-08-02
Terms and Conditions Component

BReimerTI:


packages/registration-form/package.json
```diff
@@ -20,6 +20,7 @@
     "@apollo/client": "^3.6.9",
     "@graphql-codegen/introspection": "^2.2.0",
     "graphql": "^16.5.0",
+    "html-react-parser": "^3.0.1",
```
rjschill87, line 23: Was this used anywhere or leftover from some testing?

packages/registration-form/src/components/terms-conditions-modal/terms.tsx
```diff
@@ -0,0 +1,864 @@
...
+        __typename
+        name
+      }
+    }
+  `;
+
+  const { data } = useQuery(query);
+  const companyName = data && data.CompanyDetails.name;
+  // Apply tailwind stylings to global terms
+  if (globalTerms.includes('<ol>')) {
```
rjschill87, line 20: Have you explored `html-react-parser`'s `replace` function here?

https://github.com/remarkablemark/html-react-parser#replace

My concern with this approach is that we could be erasing any attributes that may have been added previously (as unlikely as they may be)

rjschill87 (commented):
@BReimerTI Would you mind adding

```json
"publishConfig": {
    "access": "public"
}
```

to the registration form package's `package.json`? Otherwise we'll have issues w/ publishing/installing the package. Thanks!

packages/registration-form/src/components/terms-conditions-modal/terms.tsx
BReimerTI: @rjschill87 I was able to utilize the `replace` function to check for existing attributes.

rjschill87 (approved):

BReimerTI (commented):
@rjschill87 I needed to update the `registration-form` story in Storybook and move query the terms on click rather than on render.

PR #144, opened 2022-08-08
feat: use stackblitz

ljiang-ti:
Closes CLM-4455

According to Stackblitz, one can easily follow a link to start a Stackblitz project sourcing the github repository. Here is a sample link: https://stackblitz.com/fork/github/thoughtindustries/helium/tree/use-stackblitz/tooling/template-base. Once the changes are merged, ideally, the link will source from the main branch instead of the feature branch.

Encountered with some issues following the above sample link. Listed below are solutions to address these issues. Please note that not all solutions are addressed in this repo. 

## Issues

Running the helium demo app in Stackblitz encounters with the following issues:

1. Stackblitz defaults to running `dev` script defined in project's package.json file. when initializing the helium demo app, there is a prerequisite step to authentication the app. Without this step, launching the Stackblitz project would fail to run the `dev` script.
2. it seems that all network requests run in Node or browser environment of the demo app are handled by Stackblitz through a proxy to run in the browser environment. This would trigger CORS error if the servers disable CORS.
3. error occurs when running `npm run dev` command in Stackblitz following the sample link. Error message seems related to `ts-node ./server` process and the error causes the process to exit. Though the error message is hard to decipher the root cause (it reads from one of the vite's bundles `const match = /:(\d+):\d+\)$/.exec(e.stack.split('\n')[1]);...TypeError: Cannot read properties of undefined (reading 'split')`). The same error occurs when running the plain `vite-plugin-ssr` project in Stackblitz (following [this sample link](https://stackblitz.com/fork/github/brillout/vite-plugin-ssr/tree/main/boilerplates/boilerplate-react-ts))
4. file content for `locales/translations.json` is auto generated when running helium cli commands `authenticate` and `update-translations`. currently, this file is git ignored and Stackblitz will ignore the file when sourcing from the git repo. without the file, running helium cli command `dev` would trigger error for the missing file.
5. (sort of similar to the issue above) file content for `renderer/tailwind.css` is auto generated when running helium cli command `dev`. currently, this file is git ignored and Stackblitz will ignore the file when sourcing from the git repo. without the file, running helium cli command `dev` would trigger error for the missing file.

## Solutions to issues

Listed below are solutions to the above issues. Some solutions (mostly CORS related) are captured outside of this git repo and such solutions could unblock some verifications for running the Stackblitz project.

- for default Stackblitz behavior to run `dev` script, add a config file `.stackblitzrc` to override the starting command.
- for CORS related issue, would address outside of this repo. there are 2 parts: 
  - part 1: use a feature flag to allow stackblitz for helium. this would allow CORS from Stackblitz project to helium endpoint
  - part 2: re-route request during authentication to helium endpoint. the CORS will be all managed under helium endpoint. 
- for `ts-node` related issue, replace it with `tsup` and change how dev server is run.
- for git ignored `translations/localizations.json` issue, remove the file from origin and only leave the empty folder `translations` in the origin. given this change, we won't have changes to this file incurred during development/deployment to live in origin.
- for git ignored `renderer/tailwind.css` issue, update `.gitignore` to keep this file as well as any incurred changes to this file in the origin. 

## Verifications

Run the following steps from the sample link to Stackblitz project:

1. Open the sample link to launch Stackblitz project.
2. (PENDING) In the project terminal, run `npm run authenticate -- -k` to authenticate app.
  a. prerequisites: all solutions to CORS related issue
3. In the project terminal, run `npm run update-translations -- -k` to update translations.
  a. prerequisites: need a valid `ti-config.json` file and allow Stackblitz for helium endpoint
4. In the project terminal, run `npm run dev -- -k` to run dev server.
  a. prerequisites: need a valid `ti-config.json` file and allow Stackblitz for helium endpoint


ljiang-ti:
@rjschill87 @chrisvariety 

with TI QA instance deployed with the changes for "allow Stackblitz for helium", i was able to verify the following in the stackblitz project against the instance:

When feature is off:

- verify that same origin requests still works
- verify that command `npm run update-translations` fails with CORS error in stackblitz
- verify that command `npm run dev` and launching the demo app fails with CORS error in stackblitz

When feature is on:

- verify that same origin requests still works
- verify that command `npm run update-translations` succeeds in stackblitz
- verify that command `npm run dev` and launching the demo app succeeds in stackblitz

Since we will address the `re-route request during authentication to helium endpoint. the CORS will be all managed under helium endpoint.` in separate tickets, maybe we can move the verification for command `npm run authenticate` in the separate tickets?

rjschill87 (approved):
@ljiang-ti lgtm, thanks for all the extra work here! As far as re-testing the authentication flow, I agree it can be handled in a separate ticket. I went ahead and added a note to test it in CLM-5218's Notes. If there were no issues with the `update-translations` script then I'm guessing the same will hold true for `authenticate`, but no harm in double-checking.

PR #147, opened 2022-08-17
Placeholder course layout

rjschill87:
Course Layout / Topic Handling. Placeholder styling until we receive themed mockups from design.


tooling/template-base/components/CourseLayout.tsx
```diff
@@ -0,0 +1,83 @@
...
+  title?: string;
+  coursePages: Array<JSX.Element>;
+}) {
+  const pageContext = usePageContext();
+  let pageQuery = 1;
+
+  const { search } = pageContext.urlParsed;
+  if (search?.page && parseInt(search.page)) {
+    pageQuery = parseInt(search.page);
+  }
```
chrisvariety, line 18: `parseInt` gets a little wild when called without a radix, we should always specify `parseInt(foo, 10)`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

It would be ideal to avoid calling this twice as well, we can have a `isNaN` check to set it to a sane value (`1`) if provided invalid input.

tooling/template-base/components/CourseLayout.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const handleBackward = () => {
+    if (currentPageIndex > 0) {
+      setCurrentPage(currentPageIndex - 1);
+    }
+  };
+
+  useEffect(() => {
+    setNextIsDisabled(currentPageIndex >= coursePages.length - 1);
+    setPrevIsDisabled(currentPageIndex === 0);
+  }, [currentPageIndex, coursePages.length]);
```
chrisvariety, line 42: I don't think any of this needs to be state or `useEffect` at all --

`const nextIsDisabled = currentPageIndex >= coursePages.length - 1;`

https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

chrisvariety (commented):
looks like a great start!! 😍

tooling/template-base/components/CourseLayout.tsx
```diff
@@ -0,0 +1,78 @@
...
+  title?: string;
+  coursePages: Array<JSX.Element>;
+}) {
+  const pageContext = usePageContext();
+  let pageQuery = 1;
+
+  const { search } = pageContext.urlParsed;
+  if (search?.page) {
+    const pageInt = parseInt(search.page, 10);
+    pageQuery = isNaN(pageInt) ? 1 : pageInt;
```
chrisvariety, line 18: this lets negative / zero slip through. we may want to just go ahead and check for `coursePages[pageInt]` for ultimate validity?

chrisvariety (approved):
perf!

tooling/template-base/components/CourseLayout.tsx
rjschill87: Felt a little messy to do `coursePages[pageInt - 1]` cause of index differences, so opted for checking if `pageInt < 1` instead if that works.

chrisvariety (approved):

PR #150, opened 2022-08-18
feat: use local catalog error component

ljiang-ti:
Unticketed

An issue occurred when loading catalog page of the helium app. The page did quick flash from content to blank page and the console shows error about missing catalog context. Can repro this in helium app script, but not able to repro from helium template base workspace.

think the culprit might lie in how vite scanned and bundled the assets when importing from package files source directly (instead of from the package entry point). As noted the console error for missing Catalog Context in the browser, it shows the error stack from `CatalogProvider` (vite bundled asset) to `CatalogError` (not bundled by vite).

This could cause the `CatalogError` to access a different context object than the one populated by `CatalogProvider`. changes to fix this: Instead of importing from package files source directly for CatalogError (change `import CatalogError from '@thoughtindustries/catalog/src/catalog-error';`), we could clone this component in the template base.


rjschill87 (approved):
Lgtm. Appreciate you taking care of that!

PR #154, opened 2022-08-23
feat: add engine version requirements to template-base

rjschill87:
Closes CLM-5541 

Adds engine version requirements to `template-base`'s `package.json`. Since `engineStrict` property is deprecated, also added `engine-strict` to `npmrc` to flag the issue instead of only presenting warnings to user.


chrisvariety (approved):
thanks! real user feedback right here

PR #156, opened 2022-09-01
fix: graphql codegen plugin to include enum variables

ljiang-ti:
Unticketed

The custom graphql codegen plugin is used to scan documents and generate TS resources. We ran into an issue when using enum variables in the document. For some reason, the plugin does not discover the enum variables and omit the TS enums in the generated resources. The fix seems pretty straightforward. 

To verify this, include some temporary code to show the resolved issue (new workspace `packages/test`). If the changes look good, will remove the temporary code. Verify the following:
- Existing behavior: use a document with input type variables, expect that the TS resources are generated for the input types. (demo document [here](https://github.com/thoughtindustries/helium/blob/fix-graphql-codegen/packages/test/src/UpdateAssessmentAttempt.graphql))
- Issue behavior: use a document with scarlar and enum type variables, expect that the TS resources are generated for both the scarlar and enum types. (demo document [here](https://github.com/thoughtindustries/helium/blob/fix-graphql-codegen/packages/test/src/RusticiLaunchScorm.graphql))

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33339,6 +33339,10 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/test": {
+      "resolved": "packages/test",
+      "link": true
+    },
     "node_modules/test-exclude": {
       "version": "6.0.0",
       "dev": true,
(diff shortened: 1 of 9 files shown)


rjschill87 (approved):
Awesome @ljiang-ti - thanks for digging into that!

PR #163, opened 2022-09-22
Clm 5792 - create error message if deploying without helium feature flag helium repo

AlexONeillTI:


tooling/cli/lib/deploy.js
```diff
@@ -196,31 +195,43 @@ async function triggerBatch(instance, key) {
...
+          responseData = {
+            key,
+            signedUrl
+          };
+          resolve(responseData);
+        } else if (resObj && resObj.errors) {
+          const err = resObj.errors[0];
+          reject(err.message);
+        }
+        resolve(responseData);
```
rjschill87, line 229: Don't need this `resolve` here since it'll either hit `resolve` or `reject` in the `if/else`.

tooling/cli/lib/deploy.js
```diff
@@ -196,31 +195,43 @@ async function triggerBatch(instance, key) {
...
-      signedUrl
+    const options = {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ query: KEY_QUERY, variables: { nickname: instance.nickname } })
     };
-  }
 
-  return responseData;
+    let responseData = {};
```
rjschill87, line 207: Can just move this into the `if` block on L212 and define it as `const` at L220.

tooling/cli/lib/helpers/filepaths.js
```diff
@@ -3,19 +3,21 @@ const path = require('path');
...
+      const filePath = path.join(dir, content);
+      const fileStat = await stat(filePath);
 
+      if (fileStat.isFile()) {
+        newFilePaths.push(filePath);
+      } else {
+        await getFilePaths(filePath, newFilePaths);
+      }
+    }
+  } catch (err) {}
```
rjschill87, line 20: Can remove the `err` from this `catch` since we're not using it.

rjschill87 (commented):
Looks great! Couple small comments, and can remove `tailwind.css` change since it's not needed for this PR

AlexONeillTI:
Hey Rob, thanks for the review, the changes have been pushed

tooling/cli/lib/deploy.js
```diff
@@ -196,31 +195,42 @@ async function triggerBatch(instance, key) {
...
+      body: JSON.stringify({ query: KEY_QUERY, variables: { nickname: instance.nickname } })
     };
-  }
 
-  return responseData;
+    fetch(endpoint, options)
+      .then(r => r.json())
+      .then(res => {
+        const resObj = res;
+        const responseData = {};
```
rjschill87, line 211: Sorry wasn't clear; meant you could just do

```javascript
const responseData = {
  key,
  signedUrl
}
```

on L220. Think current implementation would cause an error attempting to redefine `const`

rjschill87 (approved):

PR #167, opened 2022-10-03
CLM-6033 center navbar in starter app

AlexONeillTI:


tooling/template-base/components/Navigation/UserLoginNavBar.tsx
```diff
@@ -4,68 +4,71 @@ import Logo from '../Logo/Logo';
 import icon from '../../renderer/hamburger.svg';
 import xicon from '../../renderer/xicon.svg';
 
-export default function UserLoginNavBar() {
+export default function CurrentUserNavBar() {
```
rjschill87, line 7: Was the function rename intentional? Think it may cause a little confusion since it's named the same as the func exported from `Navigation/CurrentUserNavBar`

AlexONeillTI: ah no, it wasn't my mistake! Will fix now

AlexONeillTI: error corrected and pushed

rjschill87 (approved):

PR #169, opened 2022-10-25
feat: upgrade deps

ljiang-ti:
Closes CLM-6029


tooling/template-base/renderer/_default.page.server.tsx
```diff
@@ -26,39 +23,25 @@ export const passToClient = [
...
 
-async function render(pageContext: PageContext) {
-  const { pageHtml } = pageContext;
+export { render };
+
+type RenderFn = (pageContext: PageContext) => Promise<{
+  documentHtml: ReturnType<typeof escapeInject>;
+  pageContext: Record<string, any>;
+}>;
+const render: RenderFn = async pageContext => {
```
ljiang-ti, line 32: Got TS error `Return type of exported function has or is using name 'TemplateWrapped' from external module "/Users/lu.jiang/Documents/_src/helium/tooling/template-base/node_modules/vite-plugin-ssr/dist/cjs/node/html/renderHtml" but cannot be named.ts(4058)`.

Used TS utility to explicitly type the render function instead.

tooling/template-base/renderer/_default.page.server.tsx
```diff
@@ -73,8 +56,26 @@ async function onBeforeRender(pageContext: PageContext) {
     </ApolloProvider>
   );
 
-  const pageHtml = await renderToStringWithData(App);
+  const pageHtml = await getDataFromTree(tree);
   const apolloInitialState = apolloClient.extract();
 
-  return { pageContext: { pageHtml, apolloInitialState, documentProps, queryParams, authToken } };
-}
```
ljiang-ti, line 80: The original implementation returns additional variables for `pageContext` (like `documentProps`, etc), not sure if these need to be ported over too.

tooling/template-base/types.ts
```diff
@@ -44,25 +44,27 @@ export type CurrentUser = {
 };
 
 export type PageContext = {
+  _pageId: string;
   Page: (pageProps: PageProps) => React.ReactElement;
   pageProps: PageProps;
-  pageExports: {
+  exports: {
```
ljiang-ti, line 50: the updated `vite-plugin-ssr` deprecated several fields on the page context, including `pageExports`, `url`, `urlParsed.searchString`. These fields are changed to `exports`, `urlOriginal`, `urlParsed.searchOriginal` respectively. updated the references accordingly.

tooling/cli/package-lock.json
```diff
@@ -1,602 +0,0 @@
-{
```
ljiang-ti, line 1: not sure if the package lock files under workspaces are intended to stay. remembered earlier the lock files were generated when using `lerna` to manage workspaces. Since we moved to `npm` to manage workspaces, we have a single lock file under root. might consider cleaning up these workspace lock files? (also ones from `tooling/helium-server`, `packages/contact-block`.)

tooling/helium-server/package.json
```diff
@@ -38,20 +38,21 @@
...
-    "@vitejs/plugin-react-refresh": "^1.3.6",
+    "@apollo/client": "^3.7.0",
+    "@vitejs/plugin-react": "^2.1.0",
     "crypto-hash": "^2.0.1",
-    "express": "^4.17.3",
+    "express": "^4.18.2",
     "graphql-playground-middleware-express": "^1.7.23",
     "isomorphic-unfetch": "^3.1.0",
-    "vite-plugin-ssr": "^0.3.59"
+    "vite": "^3.1.8",
```
ljiang-ti, line 47: Without declaring the dependency for `vite`, when running `template-base` demo app in the monorepo, ran into a strange error below:

```
Cannot find module vite
from tooling/helium-server/node_modules/@vitejs/plugin-react/dist/index.cjs
```

Not totally sure about the cause tho, but it seems that `@vitejs/plugin-react` has declared dev dep for `vite` from `workspace:*`. (see reference [here](https://github.com/vitejs/vite/blob/main/packages/plugin-react/package.json#L52-L54)). Declaring `vite` as a dep here resolves the error.

tooling/helium-server/worker/index.js
```diff
@@ -65,3 +65,22 @@ function isAssetUrl(url) {
   const { pathname } = new URL(url);
   return pathname.startsWith('/assets/') || pathname.endsWith('manifest.json');
 }
+
+export default {
```
ljiang-ti, line 69: All the changes related to worker is captured in df85686407aab373de757584b10bb92cd004f00f.

Was having trouble with testing these changes. I followed the steps below:

1. upgrade wrangler to latest version v2 (this is installed globally on my machine)
2. run wrangler build in `template-base` from the monorepo (the wrangler cli changes automatically from build to `wrangler publish --dry-run --outdir=dist`)

The last step failed for me, fyr, attached here is the output  log from the last step. Some of these errors seem strange, not sure if this is due to running the command in a monorepo setup. (had similar module resolution errors occur while upgrading the deps across 2 workspaces between `helium-server` and `template-base` .)
[wrangler_error.log](https://github.com/thoughtindustries/helium/files/9864622/wrangler_error.log)

package.json
```diff
@@ -26,19 +28,20 @@
...
     "postcss": "^8.3.8",
     "prettier": "^2.4.1",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "react-i18next": "^11.14.2",
-    "storybook-addon-apollo-client": "^4.0.9",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-i18next": "^11.18.6",
+    "storybook": "^6.5.13",
```
ljiang-ti, line 41: related changes to react upgrade. latest storybook version 6.5+ accepts react 18 in their peer dependency. For some reason, the set of tools (like addons) are not 100% compatible with react v18. For example, some core addons still uses deps that does not support react v18. 

According to storybook, their official support for react v18 will be available in the next major version v7. Took a look into v7 pre-release, ran into issues with typescript...

Despite the peer dep compatibility for react v18 in storybook v6.5, i was able to do the upgrade and run storybook. It's just that storybook renders the stories with react v17. There is warning about it, also some additional warnings during `npm i` about the conflict when resolving peer dep for react (from core addons).

Example warning in the console when running storybook:
```
ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it’s running React 17. Learn more: https://reactjs.org/link/switch-to-createroot
```

What are your thoughts about this? We could wait till storybook v7 is stable to upgrade this.

ljiang-ti (commented):
this is an initial draft of changes. Took most of the changes from cloud academy `cognition` branch, with some exceptions. I didn't copy all the changes related to course pages and react mdx layouts from cloud academy, thought these would be useful for future reference.

When running tests to verify the changes, ran into some issues which might relate to monorepo and module resolution. One of the major issues was about `Invalid hook call.` when loading a page that uses `useQuery` or `useMutation` hooks. For some reason, the `@apollo/client` is hoisted to the root and react v17 is installed at the root too. Since the template-base uses react v18 and detects react v17 at the root referenced by `@apollo/client`, it triggers the error. Went ahead and upgrade react as well in the monorepo.

the one remaining item is to update worker. captured some questions regarding the approach. please also see some comments inline for additional info.

tooling/template-base/package.json
```diff
@@ -11,15 +11,15 @@
     "dev:server": "tsup ./server/index.ts && node ./dist/index.js",
     "build": "npm run build:css && npm run build:vite && npm run build:worker",
     "build:css": "tailwindcss -o ./renderer/tailwind.css",
-    "build:vite": "vite build && vite build --ssr",
+    "build:vite": "vite build",
```
ljiang-ti, line 14: got warning when running dev server `[vite-plugin-ssr@0.4.43][Warning] The CLI call `$ vite build --ssr` is superfluous since `$ vite build` also builds the server-side. If you want two separate build steps then use https://vite-plugin-ssr.com/disableAutoFullBuild or use Vite's `build()` API.`

will remove this part of the script.

tooling/helium-server/package.json
```diff
@@ -14,6 +14,7 @@
       "import": "./dist/index.mjs",
       "types": "./dist/index.d.ts"
     },
+    "./worker": "./worker/index.js",
```
ljiang-ti, line 17: please remove.
```suggestion
```

ljiang-ti (commented):
some updates related to worker issue

packages/cart/package.json
```diff
@@ -22,17 +22,19 @@
   "bugs": {
     "url": "https://github.com/thoughtindustries/helium/issues"
   },
-  "dependencies": {
+  "devDependencies": {
```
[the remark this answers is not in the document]
ljiang-ti: discussed w/ @rjschill87 offline, pushed fixes to declare these as `dependencies` instead of `devDependencies`.

tooling/helium-server/worker/index.js
ljiang-ti: closing this thread.

tooling/cli/lib/deploy.js
```diff
@@ -334,3 +342,11 @@ function wait(ms = 1000) {
...
   });
 }
+
+function getWorkerVersion() {
+  const { vite: viteVersion } = packageManifest.dependencies || {};
+  if (semver.satisfies(semver.coerce(viteVersion), '>=3')) {
+    return 'v2';
+  }
+  return 'v1';
+}
```
ljiang-ti, line 352: determine worker version based on the `vite` version specified in package.json file. there might be a better way to check for the worker version.

ljiang-ti (commented):
starting this new thread to discuss the deployment process and potential changes. it seem that the current deployment flow for the helium app is: run helium cli deploy script in helium app -> create archive for current helium app directory and upload to aws s3 -> sends request to helium endpoint to trigger aws batch job -> polls the aws batch job until it completes. The aws batch job downloads the archive from aws s3, then handles the cloudflare/wrangler configuration to publish.

Currently the aws batch job uses wrangler cli v1. Looks like we could use wrangler cli v2 with the major vite version upgrade (referring to vite-plugin-ssr's cloudflare example). Discussed with @rjschill87 that we could make the following changes to allow the deployment flow to support both major versions of wrangler cli:

- update helium deployer (a docker container) the aws batch job runs, in order to handle worker version detection and configuration, PR: https://github.com/thoughtindustries/iron-workers/pull/170

For references only, we decided to use the current approach to simplify changes and maintenance for future iterations. Below was the earlier approach that would involve multiple services:
<details>
  <summary>Click to expand earlier approach</summary>
  
  Changes will involve passing through a worker version detected from helium cli to the helium deployer.

- update helium endpoint to accept a new optional argument for worker version, this value will be passed to aws batch job, PR: https://github.com/thoughtindustries/ti/pull/8593
- update helium deployer (a docker container) the aws batch job runs, in order to handle different worker versions, PR: https://github.com/thoughtindustries/iron-workers/pull/170
- update helium cli deploy command to programmatically decide on the worker version and pass it to the helium endpoint
</details>

Did some sanity check on the deployment and added notes in PR for helium deployer. would appreciate any feedbacks/comments.

tooling/template-base/renderer/_default.page.server.tsx
rjschill87: `queryParams` and `authToken` would possibly be needed as well. Believe the `catalog` component uses the `queryParams` returned from `getPageContext`, right? Think testing whether the catalog still functions as expected with the `queryParams` removed would answer this question.

tooling/cli/package-lock.json
rjschill87: @ljiang-ti Yeah, think cleaning up the workspaces here makes sense since `learn` isn't being used to manage workspaces.

package.json
rjschill87: @ljiang-ti Since the impact is fairly isolated -- and it seems like it technically still works (albeit as v17) -- I think we can leave the implemented upgrade for now, and upgrade storybook to v7 when it releases.

ljiang-ti: that sounds good to me

tooling/template-base/package.json
```diff
@@ -33,33 +33,33 @@
...
     "@types/node": "^17.0.23",
-    "@types/react": "^17.0.39",
+    "@types/react": "^18.0.21",
     "@types/react-dom": "^18.0.6",
     "tsup": "^6.2.1",
     "typescript": "^4.5.5"
   },
   "engines": {
-    "node": ">=14.17.6",
+    "node": ">=14.18.0",
```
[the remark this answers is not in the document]
ljiang-ti: more version bumps are addressed in changes to support wrangler cli v2

tooling/template-base/renderer/_default.page.server.tsx
ljiang-ti: thanks for the inputs! after reading more on vite-plugin-ssr, if i am understanding this correctly, how page context works is as follows:

- page context is bound to server by default
- server initializes page context (in helium, page context is initialized with current user, auth token, apollo client, etc)
- expose server page content to client by hook `passToClient` in `*.page.server.js`
- customize server render behavior by hook `render` in `*.page.server.js` (in helium, it exposes additional page context `apolloInitialState`)

Currently, the `passToClient` hook is implemented to expose `authToken`, `urlParsed`, etc to the client. In `catalog` page, it refers to `urlParsed` from the page context to get url pathname and extract query params. It seems to be working as is. Here is one deployed helium app with the current changes: [sample catalog page with `urlParsed`](https://v2-standalone-app-w-localnpmr.lu-jiang.workers.dev/catalog?labels=%5B%22Learning+Preference+%22%5D&values=%5B%22Blog%22%5D). think it is good for now just to expose the additional `apolloInitialState` in the `render` hook?

rjschill87: @ljiang-ti Yeah that tracks. I'm having issues viewing the sample provided though (cert error) -- are you able to deploy that against a live instance, e.g., `https://lujiang.thoughtindustries.com`

ljiang-ti: discussed w/ @rjschill87 offline, we will target staging environment to verify this through the normal helium deployment flow.

rjschill87 (approved):

PR #170, opened 2022-11-28
Update __gitignore to not include tailwind

jackanticoti:
This will make collaboration between devs easier

### DIFF
--- a/tooling/template-base/_gitignore
+++ b/tooling/template-base/_gitignore
@@ -8,5 +8,4 @@ node_modules
 .DS_Store
 dist
 *.local
-renderer/tailwind.css
 ti-config.json


jackanticoti:
@rjschill87 does this make sense to you? I've had some issues with this and figured it would make collaboration easier

rjschill87:
@jackanticoti Not necessarily opposed to the change, but what issues did you have? Can you expand on the use case of making collaboration easier? `/renderer/tailwind.css` is an auto-generated file -- it would be written with any needed styles any time a user runs `npm run dev`.

jackanticoti:
When I clone a Helium repo and run `npm install` `npm run authenticate` and then `npm run dev` I get an `Error: Could not resolve './tailwind.css' from renderer/PageWrapper.tsx` so it does not auto generate for me @rjschill87

rjschill87:
@jackanticoti how are you cloning the repo? There's a `renderer/tailwind.css` file included in `template-base` so you shouldn't be getting that error.

ooloostudio:
Hey @rjschill87 having the same problem.
When try and work on different themes I am not able to jump between different helium sites and work on them with out manually coping in the tailwind.css into the next helium site I am working on. Would be nice if the files stay same as they are on github so I can collaborate with Jack as we jump between helium demo sites. Hopefully this makes sence.

rjschill87:
@ooloostudio How/when were these projects created? Earlier versions had omitted it, but initializing new Helium apps should already include the `tailwind.css` file. It's possible you've got an older version of the `create-helium-app` cached. If you run `npx create-helium-app@latest`, do you still have this issue?

PR #172, opened 2022-12-16
feat: complete signin page

ljiang-ti:
Closes CLM-5968

It seems that there are 3 main components in the helium app: helium graphql endpoint, helium app server, helium app client. User related data (like current user and auth token) is handled from the helium app server side, then the data is hydrated to helium app client side. 

When running helium app in dev environment, the helium app server is an express app which provides a graphql route to proxy all requests to the helium graphql endpoint, a graphiql route for graphql playground, and a catch-all route to handle server-side data fetching to compose page context (some context will be hydrated to helium client side).

When running helium app in production environment (usually the deployed helium app), the helium app server is a cloudflare worker instance. Since user will visit the same TI instance domain, requests are first handled by the core platform to handle server-side data fetching, then the same requests are proxied to the cloudflare worker instance with the server-side data attached in the request header. 

The changes here mainly focus on the helium app in dev environment, since the helium app server behavior is already handled in production environment.

### Changes

- wire up `Login` mutation in sign in page
- changes for helium app in dev environment
  - in proxy `/graphql` route, parse response cookie for `authToken` if set from helium graphql endpoint, then pass-thru the cookie.
  - in catch-all route, parses request cookie header for `authToken` if set, then fetch current user based on the cookie, to initialize page context with the current user and auth token.
  - in proxy `/graphql` route, parse request cookie header for `authToken` if set, then compose the request made to helium graphql endpoint

### Testings

Did some sanity check for the sign-in flow, the general flow is as follows:
1. in incognito window, open the sign in page in helium app
2. sign in with valid user credentials, expect redirection to home page, cookie `authToken` is set in browser, content on home page is specific to the signed-in user.
3. reload home page or other pages (dashboard, catalog), expect that content on the page is specific to the signed-in user.

Verified the above flow for helium app in dev environment and for deployed helium app against TI staging (need query param `preview=<school_preview_pwd>` in TI staging to bypass `school is closed` check).

Also verified the helium app graphiql playground, add `authToken` in request header for user-specific requests. This is changed from the old behavior, where all requests were sent using the user identity of test user email from the ti config.


tooling/helium-server/src/server/server.ts
```diff
@@ -30,43 +34,71 @@ export default async function setupHeliumServer(root: string, viteDevServer: any
...
+      const { body: reqBody, headers: reqHeaders } = req;
 
-      if (Array.isArray(reqBody)) {
-        reqBody.push({ user: tiInstance.email });
-      } else {
-        reqBody.user = tiInstance.email;
+      // parse request header and pass-thru authToken header
+      const reqAuthToken =
+        reqHeaders[COOKIE_OR_HEADER_NAME_AUTHTOKEN] ||
+        reqHeaders[COOKIE_OR_HEADER_NAME_AUTHTOKEN.toLowerCase()];
```
ljiang-ti, line 42: in `@apollo/client` v3.6.9, the custom request header `authToken` set in link context is transformed to lowercase. may not need to handle the lowercase check here in v3.7+

tooling/template-base/components/Signin/SigninPage.tsx
```diff
@@ -1,12 +1,29 @@
-import React, { useState } from 'react';
+import React, { useState, FormEvent } from 'react';
+import { gql, useMutation } from '@apollo/client';
 import Logo from '../Logo/Logo';
 
+const LoginDocument = gql`
```
ljiang-ti, line 5: use inline graphql document here. 

fyr, also tried using the `useLoginMutation` from `@thoughtindustries/user`, ran into some strange server-bundle issue. The error `ReferenceError: window is undefined` only occurs to the deployed helium app and renders the `signin` page with 500 error. It seems that the module `html-dom-parser` could be the cause, which offers bundles for both browser and server environments. When built with wrangler v1 and webpack, the worker bundle is built with the browser version of the module (default behavior when webpack targets `webworker`, that it refers to the `browser` field of the module.) couldn't figure out a solution to this yet.

ljiang-ti (commented):
sorry for the large changeset, did a merge from latest staging before starting the work, my work can be viewed from this link (selected commit range) https://github.com/thoughtindustries/helium/pull/172/files/f4e66a42d1c9346fcb2ac2c91da4cc8d147e4003..19b653f854659bc7d559c3d60b31f71bf38db65b

please see inline some additional notes.

tooling/template-base/components/Signin/SigninPage.tsx
rjschill87: Note: discussed offline; going to handle update of `@thoughtindustries/user`'s `useLoginMutation` in a separate ticket, as there is a likely possibility of larger changes that would be out of scope for this ticket.

tooling/template-base/components/Signin/SigninPage.tsx
```diff
@@ -1,12 +1,29 @@
...
 
-  const signInHandler = e => {
-    console.log('Login Mutation Here');
+  const signInHandler = async (e: FormEvent) => {
+    e.preventDefault();
+    await login({
+      variables: { email: userEmail, password: userPassword }
+    })
+      .then(() => {
+        window.location.href = '/';
```
rjschill87, line 22: Bit of a nitpick since users can change this to whatever location they want, but think we can copy existing functionality and redirect users to `/learn/` as seen here:

https://github.com/thoughtindustries/ti/blob/23ac2515750a0551bf419c5afac4bc46482641a0/lms/controllers/sessions.js#L76

rjschill87 (approved):

tooling/helium-server/src/server/server.ts
ljiang-ti: quick follow up: 

the latest `staging` brought in `@apollo/client` v3.7.0. Took another look into this after merging from `staging`, still no luck getting the request headers to preserve the original upper case. What I did was to add option `preserveHeaderCase: true` to all instances of `BatchHttpLink`, according to this doc:
https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http#preserveheadercase

PR #180, opened 2023-01-20
CLM-7339 Atoms Support

rjschill87:
Closes CLM-7339

Ref:
https://github.com/thoughtindustries/ti/pull/8843
https://github.com/thoughtindustries/iron-workers/pull/175

Updates cli deploy command to look for atoms resources, and if they exist, parses the generated hash from the built project and includes them in `HeliumDeploymentStatus`, where - upon successful deploy - they will be saved for reference on the company object.

### DIFF
--- a/tooling/cli/lib/deploy.js
+++ b/tooling/cli/lib/deploy.js
@@ -83,9 +83,8 @@ const JOB_QUERY = /* GraphQL */ `
     let atomsStyleHash;
     if (hasAtoms) {
       atomsScriptHash = await getAtomsHash(OP_DIR, 'script');
-      atomsStyleHash = (await getAtomsHash(OP_DIR, 'style')) || atomsScriptHash;
-
-      await compileStyles(OP_DIR, atomsStyleHash);
+      const possibleStyleHash = (await getAtomsHash(OP_DIR, 'style')) || atomsScriptHash;
+      atomsStyleHash = await compileStyles(OP_DIR, possibleStyleHash);
     }
 
     // hash queries for whitelisting
--- a/tooling/cli/lib/deploy.js
+++ b/tooling/cli/lib/deploy.js
@@ -17,6 +17,7 @@ const {
   transformDoc,
   hashQuery
 } = require('./helpers/graphql');
+const { dirHasAtoms, getAtomsHash, copyCompiledTailwind } = require('./helpers/atoms');
 
 const OP_DIR = process.cwd();
 const TMP_DIR = os.tmpdir();
@@ -41,8 +42,8 @@ const BATCH_QUERY = /* GraphQL */ `
 `;
 
 const JOB_QUERY = /* GraphQL */ `
-  query HeliumDeploymentStatusQuery($jobId: ID!) {
-    HeliumDeploymentStatus(jobId: $jobId)
+  query HeliumDeploymentStatusQuery($jobId: ID!, $key: String, $atomsHash: String) {
+    HeliumDeploymentStatus(jobId: $jobId, key: $key, atomsHash: $atomsHash)
   }
 `;
 
@@ -64,13 +65,24 @@ const JOB_QUERY = /* GraphQL */ `
     await generateTranslationFile(OP_DIR, [instance], true);
     // trim translations file down to only keys used in project
     await gatherUsedTranslations();
+    // build project
+    const hasAtoms = await dirHasAtoms(OP_DIR);
+    await buildProject(hasAtoms);
+
+    let atomsHash;
+    if (hasAtoms) {
+      atomsHash = await getAtomsHash(OP_DIR);
+      // copy over tailwind file to compiled atoms dir
+      await copyCompiledTailwind(OP_DIR, atomsHash);
+    }
+
     // hash queries for whitelisting
     await writeGraphqlManifest();
     await uploadHeliumProject(policyData);
     // reset translations file to include all keys for local development
     await resetTranslationFile();
     const batchJobId = await triggerBatch(instance, key);
-    const fetchStatus = () => checkDeploymentJobStatus(instance, batchJobId);
+    const fetchStatus = () => checkDeploymentJobStatus(instance, batchJobId, key, atomsHash);
     const processing = result => {
       if (result === 'FAILED') {
         throw new Error('Batch deployment failed');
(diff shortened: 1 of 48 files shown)


tooling/template-base/atoms/AudioPlayerBar.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { useEffect, useState } from 'react';
+
+export default function TopicAudioPlayer({ keyName }: { keyName: string }) {
+  return <AudioP keyName={keyName} />;
+}
+
+function AudioP({ keyName }: { keyName: string }) {
```
chrisvariety, line 7: There's no need for `AudioP` here, the function definition can just be inline with `TopicAudioPlayer`. (this was just me debugging some React behavior I was seeing)

tooling/template-base/vite.atoms.config.ts
```diff
@@ -0,0 +1,28 @@
...
+      entry: resolve(__dirname, 'atoms/index.ts'),
+      name: 'TiAtoms',
+      formats: ['es'],
+      fileName: 'ti-atoms.[hash]'
+    },
+    rollupOptions: {
+      output: {
+        entryFileNames: `assets/[name]-[hash].js`,
+        chunkFileNames: `assets/[name]-[hash].js`,
+        assetFileNames: `assets/[name]-[hash].[ext]`
```
chrisvariety, line 24: nitpicky, but no need for string interpolation here..

chrisvariety (commented):
this might be more of a Slack convo, but I think we need at least one additional file... if you uncomment this line: https://github.com/thoughtindustries/cloudconnectacademy/blob/ef47cb75bc7a64bf0eaa92485a73f6215a767cce/atoms/AudioPlayerBar.tsx#L3

You'll see a `dist/assets/style-e4b6c13d.css` file gets created, I think we'll need to pull that in as well? 

It's a little tricky as your current code in this PR + the TI side assumes folks are using Tailwind, which isn't necessarily correct, I'm not even sure it is guaranteed `renderer/tailwind.css` will be created if all Tailwind usage is stripped out. I think we should really have a `atomStylesUrl` which combines styles from tailwind (if it is being used) + any other CSS file(s). I haven't explored if Vite creates just one style file or multiple if you add multiple CSS imports across atoms.

rjschill87:
@chrisvariety See what you mean on folks possibly stripping out tailwind. I'll check out how vite handles multiple files and add changes.

rjschill87:
@chrisvariety Yeah, so any styles imported in the `/atoms` dir will get compiled into a single `style-${hash}.css` file. Updated the deploy script to use that file, concat a `tailwind.css` file if it exists, and send the style hash as its own value (unless there are no imported styles, in which case it will just fallback to the script hash). If that works, can make appropriate changes on the TI side.

chrisvariety:
@rjschill87 that approach sounds great!

tooling/cli/lib/helpers/atoms.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  const regex =
+    type === 'script' ? /\/?index-([a-zA-Z]|\d)+.(ts|tsx|js|jsx)$/ : /\/?style-([a-zA-Z]|\d)+.css$/;
+
+  const idx = hashedAssets.find(asset => regex.test(asset));
+
+  if (!idx) {
+    throw new Error('No Atoms index found.');
+  }
```
chrisvariety, line 26: It seems like this will throw if `type = 'style'` if that file does not exist, but at least for `style`, it's OK if that file doesn't exist.

(the error messages are out of date as well)

tooling/cli/lib/deploy.js
```diff
@@ -64,13 +75,29 @@ const JOB_QUERY = /* GraphQL */ `
...
+    const hasAtoms = await dirHasAtoms(OP_DIR);
+    await buildProject(hasAtoms);
+
+    let atomsScriptHash;
+    let atomsStyleHash;
+    if (hasAtoms) {
+      atomsScriptHash = await getAtomsHash(OP_DIR, 'script');
+      atomsStyleHash = (await getAtomsHash(OP_DIR, 'style')) || atomsScriptHash;
+
+      await compileStyles(OP_DIR, atomsStyleHash);
```
chrisvariety, line 88: The hash here seems to only represent the hash of the `style` file, it does not take into account any tailwind.css changes. That means if you deploy, add a few tailwind classes, and re-deploy, the hash will not change, which will lead to those changes not appearing. I think we'll have to calculate a new hash entirely once we `compileStyles` (these hashes are usually just md5 or sha-1 truncated to X characters)

tooling/cli/lib/helpers/atoms.js
```diff
@@ -0,0 +1,70 @@
...
+    styles += atomsStyleContents;
+  }
+
+  const heliumTailwindPath = path.join(dir, '/renderer/tailwind.css');
+  const tailwindContents = await getFileContents(heliumTailwindPath);
+  if (tailwindContents) {
+    styles += `\n${tailwindContents}`;
+  }
+
+  const updatedStyleHash = crypto.createHash('md5').update(styles, 'utf8').digest('hex').slice(6);
```
chrisvariety, line 52: apologies if I'm jumping on this too soon, but it seems like we need to return this value so we can pass it to the deployer ?

tooling/cli/lib/helpers/atoms.js
```diff
@@ -0,0 +1,70 @@
...
+  const regex =
+    type === 'script' ? /\/?index-([a-zA-Z]|\d)+.(ts|tsx|js|jsx)$/ : /\/?style-([a-zA-Z]|\d)+.css$/;
+
+  const idx = hashedAssets.find(asset => regex.test(asset));
+
+  if (!idx && type !== 'style') {
+    throw new Error('No Atoms index found.');
+  }
+
+  const idxHash = idx.match(/-(([a-zA-Z]|\d)+\.)/)[0].replace(/(-|\.)/g, '');
```
chrisvariety, line 28: `idx` could be undefined here when `type = 'style'` and there is no style file found, we may want to just return early ?

tooling/cli/lib/deploy.js
chrisvariety, line 88:
```suggestion
      atomsStyleHash = await compileStyles(OP_DIR, atomsStyleHash);
```

chrisvariety (approved):

PR #182, opened 2023-02-07
Upgrade Story Component Format

BReimerTI:


rjschill87 (commented):
Thanks @BReimerTI ! Looks good for the most part. Running through it locally, looks like storybook is having issues with the "Story" tab for each component. 

Seems to be a common enough issue though: https://github.com/storybookjs/storybook/issues/20475

So may want to hide the `storysource` add on for now?

BReimerTI:
> Thanks @BReimerTI ! Looks good for the most part. Running through it locally, looks like storybook is having issues with the "Story" tab for each component.
> 
> Seems to be a common enough issue though: [storybookjs/storybook#20475](https://github.com/storybookjs/storybook/issues/20475)
> 
> So may want to hide the `storysource` add on for now?

@rjschill87 Thanks for taking a look! Yes, I think that works until they resolve the issue. I can push those changes up.

BReimerTI:
> 

@jackanticoti what are you thoughts ☝️

BReimerTI:
> > 
> 
> @jackanticoti what are you thoughts ☝️
> 
> I will add that removing `storybook-addon-apollo-client` will break any of the stories that are using the Apollo Client within their parameters.

@rjschill87 @jackanticoti I pushed up changes related to the storybook add-on errors.

rjschill87 (approved):

PR #183, opened 2023-02-08
feat: add cli function to fetch deploy log when deployment fails

ljiang-ti:
Closes CLM-7218

## Changes

This is a draft to add cli function to fetch deploy log when deployment fails. Referring to the current structure, we handle the deploy cli command in the command module and spawns a shell to run the deploy script. Since deployment could fail at different stage and we want to only prompt user for fetching the log when batch job fails, might consider a communication between the deploy command module and spawned deploy process. 

I am using the following structure to establish communications between processes:

- parent process: the deploy cli command module (`./lib/command-modules/deploy.js`) is responsible for the cli command declaration, orchestrating spawning/closing child processes, and prompting user questions.
- child deployment process: the spawned deploy process (`./lib/deploy.js`) is responsible for running the deployment, and conditionally closing own process or sending message to the parent process when batch job fails.
- child deployment log process: the spawned deploy log process (`./lib/deployment-log.js`) is responsible for running the deployment log fetch task, streaming full log events to local file system, and closing own process once task is complete.

### Workflow when deployment fails

When deployment fails, the communication flow looks something like below:

1. parent process spawns child deployment process with ipc established
2. child deployment process sends message when deployment fails, message body includes info about TI instance and failed batch job id
3. upon receiving message from child deployment process, parent process prompts user if they want to fetch deployment log. after user answers the question, parent process closes child deployment process. If user answers 'yes', parent process spawns child deployment log process with ipc establisted and forward the message body to trigger the fetch.
4. upon receiving message from parent process, child deployment log process triggers the full log fetch and stream log events to local file system. Once the task completes, it closes own process

### Verifications

this change would only apply to the case when batch job fails. did the following sanity checks:

1. successful deployment should behave as is
2. failed deployment due to any errors before batch job fails should behave as is
3. failed batch job should prompt user to fetch deployment log
  - when user answers 'no', all processes should be closed
  - when user answers 'yes', all log events should be saved to local file system (if less than 10k events) and then all processes should be closed afterwards
  - when user answers 'yes' and a smaller batch size is used to trigger multiple graphql calls for the log fetch. Process will simultanously stream log events to local file system until the fetch completes. When the end of the list is reached, all processes should be closed.

### DIFF
--- a/tooling/cli/lib/deploy.js
+++ b/tooling/cli/lib/deploy.js
@@ -83,51 +83,47 @@ class DeploymentError extends CustomError {
     throw new Error('Could not locate instance in configuration file.');
   }
 
-  try {
-    const policyData = await getHeliumUploadData(instance);
-    const { key } = policyData;
+  const policyData = await getHeliumUploadData(instance);
+  const { key } = policyData;
 
-    if (!policyData.key || !policyData.signedUrl) {
-      throw new Error('Could not retrieve helium keys.');
-    }
-    // ensure translations are up to date
-    await generateTranslationFile(OP_DIR, [instance], true);
-    // trim translations file down to only keys used in project
-    await gatherUsedTranslations();
-    // build project
-    const hasAtoms = await dirHasAtoms(OP_DIR);
-    await buildProject(hasAtoms);
-
-    let atomsScriptHash;
-    let atomsStyleHash;
-    if (hasAtoms) {
-      atomsScriptHash = await getAtomsHash(OP_DIR, 'script');
-      const possibleStyleHash = (await getAtomsHash(OP_DIR, 'style')) || atomsScriptHash;
-      atomsStyleHash = await compileStyles(OP_DIR, possibleStyleHash);
-    }
-
-    // hash queries for whitelisting
-    await writeGraphqlManifest();
-    await uploadHeliumProject(policyData);
-    // reset translations file to include all keys for local development
-    await resetTranslationFile();
-    const batchJobId = await triggerBatch(instance, key);
-
-    const fetchStatus = () =>
-      checkDeploymentJobStatus(instance, batchJobId, key, atomsScriptHash, atomsStyleHash);
-
-    const processing = result => {
-      if (result === 'FAILED') {
-        throw new DeploymentError('Batch deployment failed', { jobId: batchJobId, instance });
-      }
-
-      return result !== 'SUCCEEDED';
-    };
-
-    await poll(fetchStatus, processing, 3000);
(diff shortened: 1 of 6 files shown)


tooling/cli/lib/deploy.js
```diff
@@ -57,68 +57,86 @@ const JOB_QUERY = /* GraphQL */ `
   }
 `;
 
+class DeploymentError extends Error {
+  #payload;
```
ljiang-ti, line 61: using a custom error to hold additional error payload here. the `payload` property contains the batch job id and TI instance object. to avoid printing the payload info at Ln 127 below, declare it as a private class member here.

ljiang-ti (commented):
included changes for some small refactor, please see comments inline

rjschill87 (commented):
Looks great! Only thing we might want to think about is how we're chunking the logs? On one hand, it seems great to just get the chunk of the log you need, then kill the process. On the other, it feels a little tedious to possibly be stuck there confirming each prompt? My initial thought was that if they answer 'Yes' to the first question, we could loop through the pagination tokens and pipe each chunk through, or concatenate the chunks for one return. Open to thoughts here though

ljiang-ti:
> Looks great! Only thing we might want to think about is how we're chunking the logs? On one hand, it seems great to just get the chunk of the log you need, then kill the process. On the other, it feels a little tedious to possibly be stuck there confirming each prompt? My initial thought was that if they answer 'Yes' to the first question, we could loop through the pagination tokens and pipe each chunk through, or concatenate the chunks for one return. Open to thoughts here though

That's a good point and probably worth some more thoughts into these large log scenarios. Currently the cli is set to fetch the max. amount of log events that aws sdk supports (like ~10k) in a single request. Have been testing cases with less than 1k log events. Was concerned that if a full log is printed to the terminal, if anything would hit the buffer size limit (maybe the Terminal scroll back buffer size?). To account for such cases, maybe consider a stream-to-save-full-log-file?

rjschill87:
@ljiang-ti Yeah the save to log file is a good idea. Maybe instead of printing to the terminal, we just gather the paginated log events, concatenate, and save to a file in the project like `deploy-log-${timestamp}`. We're already doing it w/ the translations stuff (minus the prompt), so think it'd be doable

ljiang-ti:
Sounds good! will be looking into the proposed changes.

ljiang-ti:
@rjschill87 updated PR to stream full log events to local file system, this will replace the incremental prompt during pagination.

rjschill87 (approved):
awesome work, thanks!

PR #184, opened 2023-02-09
CLM-7704 Set duplex option when using stream as body in request

rjschill87:
Closes CLM-7704

Sets `duplex` option when using a stream for request body.

Add'l Reading: https://github.com/nodejs/node/issues/46221


chrisvariety (approved):
This still doesn't really make sense to me since we're not using node's built-in `fetch`, nor are we using `ReadableStream`... but 🤷 if it fixes it, let's try it out.

PR #185, opened 2023-02-10
CLM-7401 Update Package Generated Queries

madison-kneller-ti:
Closes CLM-7401

### DIFF
--- a/packages/catalog/src/core/utilities/parse-catalog-data/types.ts
+++ b/packages/catalog/src/core/utilities/parse-catalog-data/types.ts
@@ -5,36 +5,18 @@ export type AggregationFilter = {
   value: string;
 };
 
-export enum SortField {
-  UpdatedAt = 'updatedAt',
-  CreatedAt = 'createdAt',
-  Title = 'title',
-  PublishDate = 'publishDate',
-  CourseStartDate = 'courseStartDate',
-  Relevance = 'relevance'
-}
+export type SortField =
+  | GlobalTypes.SortColumn.UpdatedAt
+  | GlobalTypes.SortColumn.CreatedAt
+  | GlobalTypes.SortColumn.Title
+  | GlobalTypes.SortColumn.PublishDate
+  | GlobalTypes.SortColumn.CourseStartDate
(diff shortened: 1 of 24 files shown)


packages/catalog/src/core/utilities/parse-catalog-data/types.ts
```diff
@@ -9,18 +9,32 @@ export enum SortField {
...
 export enum SortDirection {
   Desc = 'desc',
   Asc = 'asc'
 }
 
 export type Sort = {
-  field: SortField;
+  field: SortColumn;
   direction?: SortDirection;
 };
```
ljiang-ti, line 39: Since the graphql layer is updated to use enums for both sort field and sort direction, alternatively we might just reference the generated typescript enums from `GlobalTypes` instead of creating our own here. There are some caveats with this approach though: seems that the TI catalog widget only uses a subset of the sort columns as declared in the graphql enum (referring to [this function in catalog widget](https://github.com/thoughtindustries/ti/blob/staging/assets/javascripts/lms/components/ti-browse-catalog.js#L9) which builds the sort dropdown). Was wondering if we want to do the same here in the catalog component. cc: @rjschill87

ljiang-ti (commented):
Looks great! The catalog component is fairly complex to refactor with the graphql changes, what you've got here covered the majority part of the changes. Left a comment regarding the supported sort columns for the catalog component.

packages/catalog/src/core/utilities/parse-catalog-data/types.ts
rjschill87: @ljiang-ti Think referencing the enums from `GlobalTypes` sounds good. Not sure I'm following your concerns w/ the caveat though: the code you linked is for determining which sort options can be displayed to the user? May need you to expand on your thoughts there.

ljiang-ti (commented):
Looks great! 👍 might be just small things: the catalog package's readme file has some reference to the layout and widget id that could be removed now; and the rendering of `LoadUserLearning` in template-base uses the string literals, we could change them to typescript enums now.

ljiang-ti (approved):

PR #186, opened 2023-02-22
feat: update to graphiql2

ljiang-ti:
Closes CLM-7638

Changes to render a static html page to serve the graphiql route, the static html was ported over from graphiql's [CDN example](https://github.com/graphql/graphiql/tree/main/examples/graphiql-cdn). It's a similar approach to the current library `graphql-playground-middleware-express`.


chrisvariety:
@ljiang-ti is there a reason to do this with a CDN vs. following a different example like https://github.com/graphql/graphiql/tree/main/examples/monaco-graphql-react-vite ?

ljiang-ti:
> @ljiang-ti is there a reason to do this with a CDN vs. following a different example like https://github.com/graphql/graphiql/tree/main/examples/monaco-graphql-react-vite ?

we are considering the CDN example for minimal customization and dependency maintenance purposes. If i understand it correctly, `graphiql` package provides a collection of react components to allow customization of the user interface. After taking a look at `monaco-graphql-react-vite` example, it seems that the example is a minimal setup to integrate with monaco editor without utilizing the existing react components from `graphiql` package. I might be missing something, will there be any concerns or things amiss going with the CDN approach? I think with the CDN approach, we could do some customizations on the look&feel (like logo).

chrisvariety:
@ljiang-ti there's a few things with the CDN approach:

 - It's not secure to rely on a third party to host javascript as that javascript could change anytime. We can host the libraries ourselves to get around this.
 - However, I saw usage of things like `react@17` in that CDN example. I figure it would make more sense to use [the version of react we're using ourselves](https://github.com/thoughtindustries/helium/blob/ca817b6146146ebebc05e2f6f27c0fc6ee04e10c/package.json#L38-L39). I didn't look too deeply into the vite example but I figured it would allow us to use our own version of react, etc.

ljiang-ti:
That's a good point for the security concern. I will take a look at the approach to build the UI ourselves given the examples.

chrisvariety:
@ljiang-ti if it's too big of a deal definitely let me know, I will admit I did not look at the vite example too deeply. We can always just host the libraries ourselves, I can provide updated URLs with our own CDN.

ljiang-ti:
Will do. 👍

chrisvariety (approved):

PR #189, opened 2023-03-09
Clm 7545 cli themes 2

madison-kneller-ti:


tooling/create-helium-app/template-base-essentials
```diff
@@ -0,0 +1 @@
+/Users/madison.kneller/ThoughtIndustries/helium-folder/helium/tooling/template-base-essentials
```
chrisvariety, line 1: not sure about this ?

chrisvariety (commented):
I may be a bit disconnected from this project (just stepping in for Rob who is out), but do we really want to duplicate all of this code across the themes? Is the expectation that the themes will diverge over time? If not, it feels like there's an opportunity here to create a shared directory which both themes could pull from vs. just copy/pasting the components.

chrisvariety:
@madison-kneller-ti We could also use different git repositories if we wanted to? That's what it looks like some folks do, e.g. https://docs.astro.build/en/install/auto/#starter-templates

... but having them in the same repository is a thing as well: https://github.com/Shopify/hydrogen/tree/2023-01/templates

It looks like Hydrogen has a 'skeleton' that it uses for all shared code, then it lets the other templates overwrite + add files as necessary: https://github.com/Shopify/hydrogen/tree/2023-01/templates/skeleton

We could try either of those techniques, it's probably more up to the product team here based on whether they see the themes having a lot of overlapping code, or slowly diverging over time. If they have a lot of overlapping code, I would go with the 'skeleton' approach, otherwise I would go with the 'multiple GitHub repositories' approach.

jackanticoti:
@madison-kneller-ti @chrisvariety product team sees the themes diverging over time so whatever approach is most appropriate with that in mind

jackanticoti:
@rjschill87 @madison-kneller-ti - I've also been thinking this over (and chatted with Todd a bit) while Product does see these themes being very different, our goal is to only build one and then have partners build other ones (using our components) so it might make sense to go with the Shopify approach which also is a lot better from a documentation perspective as well as it stays under one repo

tooling/create-helium-app/template-base-essentials
madison-kneller-ti: @rjschill87 Updated the symlink to use a relative path for `template-base-essentials`, but for `template-base`, I'm a bit confused, is what you linked not the correct symlink?

rjschill87: @madison-kneller-ti You're correct! I apologize, misread the diff as replacing the existing symlink vs. creating a new one.

tooling/template-base-essentials/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "i18next": "^21.10.0",
+    "jwt-decode": "^3.1.2",
+    "node-html-parser": "^4.1.5",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-h5-audio-player": "^3.8.6",
+    "react-i18next": "^11.18.6",
+    "remark-frontmatter": "^4.0.1",
+    "remark-mdx-frontmatter": "^2.0.3",
+    "rollup-plugin-external-globals": "^0.7.1",
```
rjschill87, line 48: Can remove this dependency from this `package.json` since `atoms` isn't included here -- same for the `template-base` `package.json`.

tooling/template-base-essentials/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "deploy": "helium deploy",
+    "dev": "helium dev",
+    "update-translations": "helium update-translations",
+    "dev:vite": "concurrently \"npm run dev:css\" \"npm run dev:server\"",
+    "dev:css": "tailwindcss -o ./renderer/tailwind.css --watch",
+    "dev:server": "tsup ./server/index.ts && node ./dist/index.js",
+    "build": "npm run build:css && npm run build:vite",
+    "build:css": "tailwindcss -o ./renderer/tailwind.css",
+    "build:vite": "vite build",
+    "build:atoms": "vite build -c vite.atoms.config.ts && vite build"
```
rjschill87, line 15: Can remove this command since `atoms` isn't included -- same in `template-base`.

tooling/template-base-essentials/pages/catalog/index.page.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
rjschill87, line 1: According to the ticket, looks like we can get rid of `./pages/catalog`, `./pages/dashboard`, and `./pages/signin`, as well as the contents of the `./components` directory, seeing as we only need the `./pages/index` to return that `h1`, and none of the components are necessary for that. (Can probably remove any unused `svg`s from `./renderer` as well.

rjschill87 (approved):

rjschill87 (approved):

PR #193, opened 2023-03-14
chore: update yml for sb7

BReimerTI:


.github/workflows/storybook.yml
```diff
@@ -19,6 +19,7 @@ jobs:
         run: |
           npm ci
           npm run build-storybook-ci
+          touch .nojekyll
```
ljiang-ti, line 22:
```suggestion
          touch ./docs-build/.nojekyll
```
might place the file in the build folder?

PR #197, opened 2023-04-18
CLM-7667-Helium-Error-Boundaries

AlexONeillTI:
As a 3rd party developer who has successfully deployed my Helium app, when I visit the deployed app and it’s not working, I should see error logging on the runtime issues.

This can be don't by adding two things:
1. React ErrorBoundary on the client side
2. React Suspense on the server side


package.json
```diff
@@ -77,6 +77,7 @@
     "npm": "8.19.2"
   },
   "dependencies": {
-    "@storybook/addon-toolbars": "^6.5.16"
+    "@storybook/addon-toolbars": "^6.5.16",
+    "react-error-boundary": "^4.0.4"
```
rjschill87, line 81: Looks like this should be in `tooling/template-base` instead of root.

Will need to make the same changes over in the `hello-world` theme as well.

AlexONeillTI:
1. Moved "react-error-boundary": "^4.0.4" to be in the correct package files
2. Added the same changes to the hello world theme

AlexONeillTI:
Fix: `react-error-boundary` is now in the correct order in both `package.json` files

rjschill87 (approved):

PR #198, opened 2023-04-18
fix: append ti-config user to request body in local dev

rjschill87:
Closes CLM-8345

SSO users cannot use the `Login` mutation as no password is set, and the `/access/` endpoints in the Thought Industries platform sets auth cookies for their respective domain, so they're not included in `localhost` requests.

### DIFF
--- a/tooling/helium-server/src/server/server.ts
+++ b/tooling/helium-server/src/server/server.ts
@@ -56,6 +56,11 @@ export default async function setupHeliumServer(root: string, viteDevServer: any
       const headers: any = { 'Content-Type': 'application/json' };
       if (reqAuthToken) {
         headers[COOKIE_OR_HEADER_NAME_AUTHTOKEN] = reqAuthToken;
+      } else if (!isProduction && tiInstance?.email) {
+        // primarily for SSO-configured schools utilizing local delevopment,
+        // as the auth cookie set via Thought Industries SSO flow will be set
+        // for that domain and will not be included in requests coming from localhost
+        reqBody.user = tiInstance.email;
       }
 
       const options = {
@@ -64,6 +69,7 @@ export default async function setupHeliumServer(root: string, viteDevServer: any
         body: JSON.stringify(reqBody)
       };
 
+      console.log('>>> reqBody', reqBody);
       fetch(heliumEndpoint, options)
         /**
          * since this runs in node env, 'isomorphic-unfetch` proxies to 'node-fetch`.
--- a/tooling/helium-server/src/server/server.ts
+++ b/tooling/helium-server/src/server/server.ts
@@ -69,7 +69,6 @@ export default async function setupHeliumServer(root: string, viteDevServer: any
         body: JSON.stringify(reqBody)
       };
 
-      console.log('>>> reqBody', reqBody);
       fetch(heliumEndpoint, options)
         /**
          * since this runs in node env, 'isomorphic-unfetch` proxies to 'node-fetch`.


chrisvariety (approved):
Not sure I totally understand, the `authToken` cookie doesn't change depending on whether you use SSO or not, as far as I know it is set here & pretty much only here: https://github.com/thoughtindustries/ti/blob/6f7dcb8d7908ac8385605d5bb223ef1d9896aa92/lib/session_helpers.js#L212-L217

I haven't looked into the issue at all though (Jira ticket just links to Discord unfortunately without additional info), so I'll definitely defer to you on this!

rjschill87:
@chrisvariety Sorry, that was poor phrasing on my part. What I was trying to say was: Non-SSO users can call the `Login` mutation and have their `authToken` cookie set with the local as the domain. SSO users developing locally don't have that option, and cannot use the `/access/` endpoints of their Thought Industries instance because that instance will set the cookie for their Thought Industries domain - so the local ApolloClient doesn't include that cookie in its requests.

chrisvariety:
@rjschill87 ah, so that's typically what the `email` in `ti-instances.json` is for (avoids login entirely), but this is specifically for testing `Login` mutation locally. I'm not sure if we can make that a bit more clear in the code, e.g. only doing this when someone calls the `Login` mutation ?

rjschill87:
@chrisvariety Poor phrasing on my part again, sorry. No it's not for testing `Login` mutation, as you mentioned -- SSO doesn't work with `Login`. They weren't actually using that, I was only saying they *don't* have the option of using it to set a `currentUser` when developing locally.

But yeah you're right, it's the reason for the `email` in `ti-instances.json`. The logic for it was removed from `helium-server` during the completion of the Sign In page: https://github.com/thoughtindustries/helium/pull/172/files#diff-44f0664d9b0b88b856ea5ff74f32f919cf06ea5b2dea14a3ecf18160e9297f59L35-L38

As that went with the release of the `Login` mutation, and made the `email` in `ti-instances.json` redundant as the `currentUser` could then be set via the `Login` mutation. Except for SSO users.

So this is just giving that functionality back to SSO users developing locally.

chrisvariety:
@rjschill87 hmm I'm not sure I overall agree with using the `Login` mutation over the `email` in `ti-instances.json`, they feel fairly separate to me, but once again I'll defer to you here.

rjschill87:
@chrisvariety In a way, the removal of the code I linked kind of forced non-SSO users to use `Login` over `email` when dev'ing locally, so this change will at least make `email` the default unless they explicitly use the `Login` mutation to set a user.

Or is your issue more along the lines that using the `Login` mutation updates login event data in the TI platform, even though they're only logging in locally?

chrisvariety:
@rjschill87 `Login` mutation has the variables `email: String!, password: String!` which fundamentally doesn't apply to SSO users. Requiring them to e.g. fill in any value for `password` doesn't really make sense either, it just doesn't align with the various different ways folks gain access to the Ti platform.

A separate `Impersonate` mutation, which only takes an `email: String!`, could make sense, and aligns with how the Ti platform works, but personally I liked the elegance of the previous system where you could just implicitly impersonate someone by specifying their email address in `ti-instances.json`.

chrisvariety:
If #172 fundamentally doesn't work, we could consider reverting it as well. It seems like we should put more thought into how that works with the `email` in `ti-instances.json`, or consider removing that `email` (which now just doesn't do anything?)

chrisvariety:
We can also just ship this to unblock folks and figure the rest out later 😄

rjschill87:
@chrisvariety Yeah don't see removing `email` really being an option as there's not much of a path forward for SSO users (aside from implementing an `Impersonate` mutation); but since most of the work for supporting `email` in `ti-instance.json` still exists seems like the best option at the moment.

PR #200, opened 2023-04-20
CLM-8347 Hello world theme

AlexONeillTI:
closes CLM-8347

Update the Hello World theme to the new comps to match https://www.figma.com/file/QEiNZVARrS9TbWVD6V9esL/Hello-World-Theme?node-id=4173%3A34114&t=cyW7gNlqMHnU9BZ6-1


tooling/template-base-essentials/components/CommunityNav.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+const CommunityNav = () => {
+  const CommunityNavElement = (props: {
```
rjschill87, line 4: Don't think it needs its own `CommunityNavElement`, does it? Could just loop through an array of objects again to render that.

tooling/template-base-essentials/pages/index/index.page.tsx
```diff
@@ -9,9 +13,42 @@ const documentProps = {
...
+      <TitleAndBody
+        title="Configure your page"
+        body="Start customzing your Helium app by going to /pages/index.page.tsx to edit this page. Check out some of the resources linked below to learn more about Helium. Have a question? Get help in our Discord and see how other developers are using Helium."
+      />
+      <div className={RenderLinksContainerStyling}>
+        <RenderLinks links={BUILDING_LINKS} linkStyling={LinkStyling} title="Start building" />
+      </div>
+      {/* RenderLinks */}
+      <div className={RenderLinksContainerStyling}>
+        <RenderLinks links={RESOURCE_LINKS} linkStyling={LinkStyling} title="Start building" />
```
rjschill87, line 50: Think the title should be `Resources`

tooling/template-base-essentials/components/CommunityNav.tsx
AlexONeillTI: I have created a NavElementArray of objects, which is looped through in RenderNavElements to return the CommunityNav

tooling/template-base-essentials/components/HomepageHeader.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+const HomepageHeader = (props: {
+  image: string | undefined;
```
rjschill87, line 4: Think we can just denote the prop types as optional, e.g., `image?: string`? Not sure it makes sense to define as required then allow `undefined`; if we were to re-use this component but without one of the properties, we'd have to pass explicitly pass `undefined` (I believe).

https://kate-dev.medium.com/when-to-use-typescript-optional-property-how-is-it-different-from-declaring-property-as-undefined-2319a0ee1f07

tooling/template-base-essentials/components/RenderLinks.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+const RenderLinks = (props: {
+  links: { [s: string]: unknown } | ArrayLike<unknown> | null;
+  linkStyling: string | undefined;
+  title: string | undefined;
+}) => {
+  const LINKS_ARRAY = [];
+  if (props.links !== null) {
+    console.log(props.links);
```
rjschill87, line 10: :bomb:

tooling/template-base-essentials/components/RenderLinks.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+const RenderLinks = (props: {
+  links: { [s: string]: unknown } | ArrayLike<unknown> | null;
```
rjschill87, line 4: Think we can work on the type def here. Since the `links` prop receives an array of objects w/ `href` and `label`, we should be able to define a `type` for use here. Don't think we need `null` either.

tooling/template-base-essentials/components/RenderLinks.tsx
```diff
@@ -0,0 +1,30 @@
...
+  linkStyling: string | undefined;
```
rjschill87, line 5: Same as above for optional vs `string | undefined`. Will likely want to revisit these throughout.

tooling/template-base-essentials/components/RenderLinks.tsx
AlexONeillTI: Have now added the following declaration in Render Links props: `links: Array<{ href: string; label: number }>;`

tooling/template-base-essentials/components/RenderLinks.tsx
AlexONeillTI: I have looked over all prop validations. Let me know if this can be improved further in any way.

rjschill87 (approved):
One last comment but otherwise looks good!

rjschill87 (approved):

PR #204, opened 2023-04-26
Remove react-key warning from homepage

AlexONeillTI:


rjschill87:
@AlexONeillTI - that was my mistake; didn't mean to submit that. It's not an issue here, but for future references, may want to include additional info for the `key` beyond just its index in the array. Otherwise you'll end up with multiple `<a>` elements w/ the same key, etc.

PR #206, opened 2023-05-19
CLM-8307 ContentHeader Component

AlexONeillTI:
Closes CLM-8307

Reusable content header component to help users get up and running


packages/content/src/components/content-header/content-header.tsx
```diff
@@ -0,0 +1,94 @@
...
+          <div key={i} className="px-1">
+            {StarIconShaded}
+          </div>
+        );
+      }
+    }
+    return <>{ratingArray}</>;
+  };
+
+  const CalculateRating = (rating: number) => {
```
rjschill87, line 49: I think we can refactor the SVG elements and the math as it feels a bit disjointed at the moment. Some math in a dedicated function, some in the `StarCount` component, then some in the returned component.

It also leads to some 'inaccuracies' in the information provided. Like in the storybook component, with a rating of 36, displayed in a factor of 5, I would expect to see `1.8`. But because of the `Math.round` and then `/ 2) + '.0'` on Line 78, I see an average of `2.0` instead.

Think a cleaner approach would to have a dedicated score for returning the score as a score out of 5, (e.g., calling `calculateRating(36)` would return `1.8`;

Then in the returned component you can just display that value without the additional ` / 2) + '.0'` transformation.

For the stars piece, think we can borrow some inspiration from SO: https://stackoverflow.com/a/71563209

With the `1.8` returned from our `calculateRating` function, we can move the decimal, and pass `18%` to the `<rect>` `width` in our svg, giving us a more accurate percentage of stars filled.

Last thing -- would rename this `calculateRating`. Know it's nitpicky, but we're using camelCase for functions/variables, and PascalCase for components/Classes.

AlexONeillTI: Refactor pushed. Some bug changes so let me know if anything needs more attention!

rjschill87: @AlexONeillTI So, with the `symbol`/`mask` used in the SVG, we don't actually need to iterate through a star count in the `RenderStars` function or bother with the `full`/`half` `clipId` stuff. We can just use the `mask` to display 5 of the `#star` elements, then set the `width` of our `rect` to `props.score` to achieve the same effect. 

Mocked up an example here if you'd like to implement it: https://github.com/thoughtindustries/helium/compare/CLM-8307-ContentHeader-Component...CLM-8307-ContentHeader-Component-rs?expand=1 Hit me up if anything doesn't make sense or you've got questions on anything.

packages/content/src/graphql/queries/CourseSlug.generated.tsx
```diff
@@ -0,0 +1,73 @@
+import * as Types from '../global-types';
+
+import { gql } from '@apollo/client';
+import * as Apollo from '@apollo/client';
+const defaultOptions = {} as const;
+export type CourseGroupBySlugQueryVariables = Types.Exact<{
+  Slug: Types.Scalars['Slug'];
```
rjschill87, line 7: @AlexONeillTI think the `slug` arg should be lowercased throughout.

packages/content/src/components/content-header/content-header.tsx
```diff
@@ -0,0 +1,88 @@
...
+        {/* header */}
+        <div className="text-5xl font-bold pb-2 font-primary">{data?.CourseGroupBySlug?.title}</div>
+        {/* description */}
+        <div className="text-lg text-slate-400 pb-2.5 font-primary">
+          {data?.CourseGroupBySlug?.description}
+        </div>
+        {/* stars */}
+        {props.showStars && (
+          <div className="flex pb-8">
+            {data?.CourseGroupBySlug?.rating && <Stars score={data?.CourseGroupBySlug?.rating} />}
```
rjschill87, line 69: Seems like we're repeating a lot of the optional chaining for `data?.CourseGroupBySlug?.` throughout -- think we could refactor so that we destructure `data.CourseGroupBySlug` before this point? Know it's a nitpick but think it'd help readability in the future.

packages/content/src/components/content-header/content-header.tsx
```diff
@@ -0,0 +1,88 @@
...
+            {data?.CourseGroupBySlug?.rating && (
+              <div className="font-bold px-1">
+                {(data?.CourseGroupBySlug?.rating / 20).toFixed(1)}
+              </div>
+            )}
+            <div>({data?.CourseGroupBySlug?.ratingsCount} Reviews)</div>
```
rjschill87, line 75: Know it'll be handled on the backend, but we should probably test for a value for `ratingsCount` and fallback to a `0` so we don't end up rendering something like ` Reviews`.

rjschill87 (commented):
Last couple comments, should be pretty small!

packages/content/src/components/content-header/content-header.tsx
AlexONeillTI: Should we default to 0, or should we emit the reviews element entirely if there are no reviews?

AlexONeillTI: After pushing a fix, let me know if this is a suitable solution

rjschill87: @AlexONeillTI That still assumes we'll always get back a `ratingsCount` if we get `ratings`; feels like the `ratingsCount` text should just be dependent on the `ratingsCount` var. Plus, if the `ratings` field is `0`, we'd end up 'hiding' this display entirely, which would mean it's essentially hiding bad (but valid) reviews. As to whether it should display `0 Reviews` or hide it entirely, that's more of a product/design decision.

packages/content/stories/content-header.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+  rating: 36,
+  ratingsCount: 4
+});
+
+const mockApolloResultsFactory = () => [
+  {
+    request: {
+      query: CourseGroupBySlugDocument,
+      variables: {
+        Slug: 'test-ribbon'
```
rjschill87, line 32: Should fix the `Slug` here as well.

packages/content/src/graphql/queries/CourseSlug.graphql
```diff
@@ -0,0 +1,9 @@
+query CourseGroupBySlug($Slug: Slug!) {
```
rjschill87, line 1: @AlexONeillTI Need to fix the `$Slug` here as well. Pretty sure fixing it here and regenerating would've fixed the issue as well. Also, may want to rename the files to something more descriptive like `CourseGroupBySlug.graphql` and `CourseGroupBySlug.generated.tsx`

packages/content/src/components/content-header/content-header.tsx
```diff
@@ -68,14 +68,18 @@ const ContentHeader = (props: {
...
-          <>
-            <div className="flex pb-8">
-              <Stars score={rating} />
-              <div className="font-bold px-1">{(rating / 20).toFixed(1)}</div>
-              <div>({ratingsCount} Reviews)</div>
-            </div>
-          </>
+        {props.showStars && (
+          <div className="flex pb-8">
+            {rating ? (
```
rjschill87, line 73: @AlexONeillTI Think there's still a bit of cleanup we can do here.

1) On line 52, we can initialize `ratingsCount` with a value of `0` instead of leaving it `undefined`. That way it's only updated if we have data from `CourseGroupBySlug`, and if we don't, there's no need for the `or 0 reviews` block.

2) `ratingsCount` is still dependent on `rating` having a truth-y value in order to render, when it should only depend on if `ratingsCount` has a truth-y value. Ex: If 5 learners rate a course as `0` (unlikely but possible); since `rating` is `0`, we would end up rendering `0 Reviews`, which would cause a learner to infer that there are no reviews on the course, when really it's just a bad course.

AlexONeillTI: The ratingsCount returns 0 if there are no reviews. So I think we can just return the ratingsCount regardless? Pushed so lmk what you think!

rjschill87 (approved):
lgtm! Appreciate the changes. Note, hold on merging until https://github.com/thoughtindustries/ti/pull/9380 is deployed.

AlexONeillTI:
@rjschill87 
Added useLearningPathBySlug query. This is used if the `content-header` is used on a `learningPath` content type. If any other content type is passed, it will still use the `useCourseGroupBySlug`.
This is spread out into two different functions. This is because the rules of react hooks do not allow hooks to be used conditionally. So I was unable to just have a condition to call the queries based on the `ContentType`.

rjschill87:
@AlexONeillTI Makes sense -- mind adding to the storybook to cover the LP too?

AlexONeillTI:
> @AlexONeillTI Makes sense -- mind adding to the storybook to cover the LP too?

Sorry, meant to mention that I was working on it! Just about to finish it up now!

packages/content/README.md
```diff
@@ -42,3 +42,30 @@ if (data) {
   });
 }
 ```
+
+## Content Header
+
+The `Content Header` component is used to add a title, description, rating and image to a course quickly. This is done by passing props to a component that is already created, rather than creating a new one from scratch.
```
rjschill87, line 48:
```suggestion
The `Content Header` component is used to display the `title`, `description`, `rating`, `ratingsCount` and image of a Course or Learning Path.
```

packages/content/README.md
```diff
@@ -42,3 +42,30 @@ if (data) {
...
+    <ContentHeader contentKind={contentKind} slug={courseSlug} showStars={true} showImage={true} />
+  );
+}
+```
+
+## Props
+
+| Name        | Required | Type                 | Description                                                    |
+| ----------- | -------- | -------------------- | -------------------------------------------------------------- |
+| contentKind | Yes      | <code>string</code>  | The contentKind of the Content. LearningPath, Course elements. |
```
rjschill87, line 68:
```suggestion
| contentKind | Yes      | <code>string</code>  | The contentKind of the Content, e.g., `learningPath`, `course`. |
```

packages/content/README.md
```diff
@@ -42,3 +42,30 @@ if (data) {
...
+| slug        | Yes      | <code>slug</code>    | Content Slug cart.                                             |
```
rjschill87, line 69:
```suggestion
| slug        | Yes      | <code>slug</code>    | The slug of the content that should be displayed.                                             |
```

packages/content/README.md
```diff
@@ -42,3 +42,30 @@ if (data) {
...
+| showStars   | No       | <code>boolean</code> | If false, hide the stars. If true, show the stars              |
```
rjschill87, line 70:
```suggestion
| showStars   | No       | <code>boolean</code> | Controls whether a course's rating is displayed, shown in stars.              |
```

packages/content/README.md
```diff
@@ -42,3 +42,30 @@ if (data) {
...
+| showImage   | No       | <code>boolean</code> | If false, hide the image. If true, show the image              |
```
rjschill87, line 71:
```suggestion
| showImage   | No       | <code>boolean</code> | Controls whether a piece of content's Detail Image is displayed.              |
```

packages/content/README.md
```diff
@@ -42,3 +42,30 @@ if (data) {
   });
 }
 ```
+
+## Content Header
+
+The `Content Header` component is used to add a title, description, rating and image to a course quickly. This is done by passing props to a component that is already created, rather than creating a new one from scratch.
+
+### Example code
```
rjschill87, line 50:
```suggestion
## Example code
```

packages/content/README.md
```diff
@@ -42,3 +42,30 @@ if (data) {
   });
 }
 ```
+
+## Content Header
```
rjschill87, line 46: Would make the below changes in https://github.com/thoughtindustries/ti-docs/pull/18 as well.

rjschill87 (commented):
Should be the last round!

AlexONeillTI:
@rjschill87 Just to put a pin on it, you're happy with the most up-to-date changes here?

PR #207, opened 2023-06-01
CLM-8779-Replicate-Redirect-on-Dashboard-Page

AlexONeillTI:
Expose redictTo from the pageContext in the ssr worker. Check if it is defined and if so, redirect to that url instead.

### DIFF
--- a/tooling/helium-server/worker/ssr.js
+++ b/tooling/helium-server/worker/ssr.js
@@ -46,13 +46,18 @@ async function handleSsr(url, authToken = null, userAndAppearanceToken = null) {
     sha256,
     authToken
   );
-  const { httpResponse } = pageContext;
+
+  const { httpResponse, redirectTo } = pageContext;
 
   if (!httpResponse) {
     return null;
   } else {
     const { statusCode, body } = httpResponse;
     const headers = assembleHeaders(pageContext);
+
+    if (redirectTo) {
+      url = redirectTo;
+    }
 
     return new Response(resolveAssetUrls(url, body), {
       headers,


tooling/helium-server/worker/ssr.js
```diff
@@ -46,14 +46,19 @@ async function handleSsr(url, authToken = null, userAndAppearanceToken = null) {
...
     return null;
   } else {
     const { statusCode, body } = httpResponse;
     const headers = assembleHeaders(pageContext);
 
+    if (redirectTo) {
+      url = redirectTo;
+    }
+
     return new Response(resolveAssetUrls(url, body), {
```
chrisvariety, line 62: not sure resolveAssetUrls makes sense when we're redirecting?

tooling/helium-server/worker/ssr.js
```diff
@@ -46,14 +46,19 @@ async function handleSsr(url, authToken = null, userAndAppearanceToken = null) {
...
+  const { httpResponse, redirectTo } = pageContext;
 
   if (!httpResponse) {
     return null;
   } else {
     const { statusCode, body } = httpResponse;
     const headers = assembleHeaders(pageContext);
 
+    if (redirectTo) {
+      url = redirectTo;
```
chrisvariety, line 59: should we set a status code as well? typically 301 or 302 I believe.

tooling/helium-server/worker/ssr.js
AlexONeillTI: i created a default status code if one is not provided. I have also set the body in the case of the redirection

tooling/helium-server/worker/ssr.js
```diff
@@ -46,14 +46,30 @@ async function handleSsr(url, authToken = null, userAndAppearanceToken = null) {
...
-    const { statusCode, body } = httpResponse;
+    let { statusCode, body } = httpResponse;
     const headers = assembleHeaders(pageContext);
 
+    if (!statusCode) {
+      //set default if code not provided
+      statusCode = 302;
+    }
+
+    if (redirectTo) {
```
rjschill87, line 63: @AlexONeillTI Think we'll want to move the check for `redirectTo` before the check for `httpResponse`. Looking at `vite-plugin-ssr`'s example, looks like there's the possibility we could get a `redirectTo` back even without an `httpResponse`, and as it's currently written, this order would miss the redirect.

tooling/helium-server/worker/ssr.js
```diff
@@ -46,14 +46,30 @@ async function handleSsr(url, authToken = null, userAndAppearanceToken = null) {
     sha256,
     authToken
   );
-  const { httpResponse } = pageContext;
+
+  const { httpResponse, redirectTo } = pageContext;
+
+  let { statusCode, body } = httpResponse;
```
rjschill87, line 52: So we don't want to try to destructure `httpResponse` here since the check on line 70 tells us `httpResponse` could be something other than an object. Think we'd be fine to leave the destructuring of `httpResponse` where it was, and just hardcode the `302` into the Response.

tooling/helium-server/worker/ssr.js
```diff
@@ -46,14 +46,30 @@ async function handleSsr(url, authToken = null, userAndAppearanceToken = null) {
...
+    statusCode = 302;
+  }
+
+  if (redirectTo) {
+    body = `Redirecting to ${redirectTo}.`;
+
+    return new Response(body, {
+      headers,
+      status: statusCode,
+      url: redirectTo
```
rjschill87, line 66: Not sure `url` is a valid prop for the `options` obj when constructing a new `Response`. Any opposition to doing something like `return Response.redirect(redirectTo, 302)`?

AlexONeillTI: I have pushed this, but is would we need to make it a `new Response...`

rjschill87 (approved):

PR #211, opened 2023-06-23
fix: catalog component malfunctioning

ljiang-ti:
Closes CLM-8876

## Cause

Catalog component uses Tailwind visibility utility classes to toggle the display for dropdown menu. The ones used are `visible` and `invisible`. When the visibility utility is disabled in tailwind config, tailwind will not generate any classes for this utility. This would cause the dropdown menu to be unresponsive in catalog component. The reason to disable the visibility utility is to avoid class name conflicts between tailwind and TI core platform (for example, the `collapse` class name), this conflict applies more likely to a TI core platform page with Atoms deployed currently. The conflict is not a concern for helium page.

## Solution

To make sure catalog component to continue function on helium pages, will update the tailwind preset package to add custom utilities for `visible` and `invisible` matching ones from Tailwind. By leaving out the `collapse`  by this approach, the rendered tailwind styles will skip this class.

Additional changes include:
- remove references to class names `row` and `collapse`, these were ported over by mistake
- fixed some more hydration errors during verification

### DIFF
--- a/packages/catalog/__tests__/catalog.test.tsx
+++ b/packages/catalog/__tests__/catalog.test.tsx
@@ -849,8 +849,7 @@ describe('@thoughtindustries/catalog', () => {
                               <strong>
                                 Course
                               </strong>
-                              | 
-                              Test Author
+                              | Test Author
                               <p
                                 class="mb-1 text-gray-700"
                               >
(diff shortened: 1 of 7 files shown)


rjschill87 (approved):
lgtm!

PR #212, opened 2023-06-24
fix: use exact vite versions

ljiang-ti:
unticketed

### DIFF
--- a/packages/catalog/package.json
+++ b/packages/catalog/package.json
@@ -31,10 +31,10 @@
     "dayjs": "^1.10.8",
     "graphql": "^16.6.0",
     "i18next": "^21.10.0",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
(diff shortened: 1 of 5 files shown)


packages/catalog/package.json
```diff
@@ -31,10 +31,10 @@
     "dayjs": "^1.10.8",
     "graphql": "^16.6.0",
     "i18next": "^21.10.0",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
+    "react": "^17 || ^18",
+    "react-dom": "^17 || ^18",
```
ljiang-ti, line 35: as a note, running version updates from the root automates these changes

ljiang-ti (commented):
verified changes in test, storybook and helium templates. the changes have some side effects, please see comments below

ljiang-ti:
Need to be revisited. Close for now.

PR #213, opened 2023-06-24
fix: cart issues

ljiang-ti:
Closes CLM-8820

changes:
- set cart cookie to path `/`
- override default encoder in `universal-cookie` when serializing cart cookie value
- currently the serialization/parsing of cookie value is handled by `window.atob` and `window.btoa`, which seems to work fine except for the handling of non-binary data. Replace it with `buffer` package to simulate node's `Buffer` API for browser.

verifications: deployed a test helium app along with some tweaks to the staging environment. The helium app only contains a dynamic path for '/courses/@courseSlug' which handles the cart rendering for ILT contents. Setup some ILT contents for purchase in staging. Verify the cart to be synced between the helium page and non-helium page. 

notes: helium feature flag is turned of in staging to avoid testing conflicts. one can turn on the feature flag to review the changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38997,17 +38997,41 @@
       "dependencies": {
         "@headlessui/react": "~1.5.0",
         "@thoughtindustries/content": "^1.2.0-beta.3",
+        "buffer": "^6.0.3",
         "clsx": "^1.1.1",
         "couponable": "^7.0.0",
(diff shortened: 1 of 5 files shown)


AlexONeillTI:
Believe this is related to CLM-8820

PR #220, opened 2023-07-19
CLM-8914 Cart Translations Issue

AlexONeillTI:
Closes CLM-8914

Issue: Cart translations were not working in a deployed environment, but did work locally. [Video of issue](https://www.loom.com/share/550479a2730f490eb2e22cb8bb136b98?sid=4c3211db-f7f1-499d-8ae6-09463158d731)

The issue is with the `i18nObject`, as it was not picking up all translations. The issue is resolved by using the `sourceDefaultNamespace` to map translations instead. 

I tested this using the `Cart` component as detailed in the ticket, but also tested the `Catalog` component to ensure that the translations were still working for other helium components. All looks good in testing.

Please let me know if there is any reason why we should use `i18nObject` instead of `sourceDefaultNamespace`.

### DIFF
--- a/tooling/cli/lib/parse-translations.js
+++ b/tooling/cli/lib/parse-translations.js
@@ -34,13 +34,10 @@ const KEYS_WITH_PLURALS = [];
 
   gatherPluralizedKeys();
 
-  const i18nStore = parser.get();
-  // en is default lang and lms is default namespace, so all keys used should be in that object
-  const i18nNSObject = i18nStore.en && i18nStore.en.lms ? i18nStore.en.lms : {};
-  const usedTranslationKeys = Object.keys(i18nNSObject);
+  const usedTranslationKeys = Object.keys(sourceDefaultNamespace);
 
   for (const key of usedTranslationKeys) {
-    const translationValue = i18nNSObject[key];
+    const translationValue = usedTranslationKeys[key];
 
     // while the keys are saved in TI files as home.meta-title, the parser will automatically
     // turn that to `home: { 'meta-title': translationValue }`


chrisvariety (commented):
any way we can add automated test coverage for this? It's fairly critical that translations work, and it sounds like they weren't properly prior to this fix, so it would be good to ensure we don't regress here in the future. I think if we separated out the steps of reading/writing files we could test only the processing step, e.g.

```
const translations = await readTranslations();
const result = processTranslations(...); // this is what gets tested
await writeTranslations(result);
```

tooling/cli/lib/parse-translations.js
```diff
@@ -34,13 +34,10 @@ const KEYS_WITH_PLURALS = [];
 
   gatherPluralizedKeys();
 
-  const i18nStore = parser.get();
-  // en is default lang and lms is default namespace, so all keys used should be in that object
-  const i18nNSObject = i18nStore.en && i18nStore.en.lms ? i18nStore.en.lms : {};
-  const usedTranslationKeys = Object.keys(i18nNSObject);
+  const usedTranslationKeys = Object.keys(sourceDefaultNamespace);
```
rjschill87, line 37: @AlexONeillTI if we're assigning `sourceDefaultNamespace` to `usedTranslationKeys`, it'd seem like we're including the entire translation file rather than the translation keys used in the actual project. That seem accurate?

AlexONeillTI: Yes it does actually, but i18nObject is only picking up the following:
`i18nStore:  {
  en: {
    lms: {
      null: '',
      '(empty string)': '',
      finished: '',
      cancelled: '',
      div: '',
      img: '',
      li: '',
      a: ''
    }
  }
}`

AlexONeillTI:
> any way we can add automated test coverage for this? It's fairly critical that translations work, and it sounds like they weren't properly prior to this fix, so it would be good to ensure we don't regress here in the future. I think if we separated out the steps of reading/writing files we could test only the processing step, e.g.
> 
> ```
> const translations = await readTranslations();
> const result = processTranslations(...); // this is what gets tested
> await writeTranslations(result);
> ```

@chrisvariety Good idea, should I include this in the scope of this ticket, or create a new ticket to handle this?

chrisvariety:
@AlexONeillTI up to you and @rjschill87, my vote would be to include it in this ticket as we generally prefer to ship code changes with corresponding tests.

tooling/cli/lib/parse-translations.js
rjschill87: @AlexONeillTI Believe the `dist` path is correct -- can run `npm run build` to verify that's where the project is compiled. Based on the output of `i18nObject` though, looks like it's possible there's a key or something with invalid characters, causing html tags to get picked up as keys instead. Would log `fileContents` there and see what that looks like.

AlexONeillTI:
@rjschill87 
After inspecting the compiled files, the translations are not only prefixed with t, but also 'o', 'p', 'f', and 'i18n' for hydratedContent. I have added these to the `parser.parseFuncFromString` call. 

I have also added the testing that @chrisvariety has requested. This involved splitting the code into three segments:
1. Reading the data
2. Processing the translations
3. Writing the data

If the translations exist, FINAL_TRANSLATIONS is defined and contains all the necessary translations. `writeTranslations` then writes the results to the translations file. 

If the translations do not exist, then FINALTRANSLATIONS = {}, which now throws a new error: "Error('No translations exist!');" 

I have tested these changes on https://alexoneilsandbox.thoughtindustries.com/cart.

AlexONeillTI:
@rjschill87 
Pushed a fix for the former part of https://thoughtindustries.atlassian.net/jira/software/c/projects/CLM/boards/26?modal=detail&selectedIssue=CLM-8914.

The translation keys must be translated directly, rather than passed as a variable. I have moved `localizedSortMapping` into `sort-selector.tsx`, as it needs to use the `useTranslation` hook within a react comp. 

Will I leave the `constants.tsx` file as is, or remove it entirely?

rjschill87:
@AlexONeillTI If that was all that was in the `constants.tsx` file and it's not referenced anywhere else, you can remove it

AlexONeillTI:
@rjschill87 Tests are now successfully passing the Github action. Let me know if the coverage of these tests is sufficient.

Note - I added a catch to filePaths.js as there was no error handling in this file.

tooling/cli/dist/server/chunks/chunk-7df435e3.js
chrisvariety, line 1: doesn't seem like we want these files committed..

AlexONeillTI: So for this ticket I had to create a "sample project" to generate the translations - as a compiled project is what parse-translations tests as can be seen here: https://github.com/thoughtindustries/helium/blob/2c94bda3bdc535b8ce772b767c3085fe6aa379b8/tooling/cli/lib/parse-translations.js#L10

The sample project is the helium starter-app, which only contains a handful of translations, which made for manageable testing.

If you would like me to remove the way I have used this compiled project, is there a more efficient way to provide `parse-translations.js` has what it needs to be tested thoroughly?

chrisvariety: not sure I totally understand the specifics here so I'll defer to @rjschill87 on that, overall we should keep compiled files out of source control.

AlexONeillTI: Okay cool, I'll circle back with @rjschill87 now so! Thanks!

AlexONeillTI:
@rjschill87 We discussed isolating the testing to evaluate the output of `npm run update-translations`

`npm run update translations` does two things:
1. Fetches the translations for the first instance in `ti-config.json`
2. Writes ALL the translations to `locales/translations.json`

To me, this doesn't seem to match the original test coverage, which was based around the processing (including reading and writing to translations.json) of the translations so we can ensure that only the used translations are deployed in the compiled project.

So my question is:
Are we now solely testing the fetch of translations from an instance (which is the testing of npm run update-`translations.json`)
Are we testing the processing of translations (the current implementation but possibly moved to `template-base`)
Or are we doing both (essentially testing the entire translation operation of the `deploy.js` script)?

rjschill87:
@AlexONeillTI You're right. I think ideally we'd want to test both, so think the approach would be to have a test that runs `npm run update-translations` that mocks the request/response for translations, so we can verify the initial `locales/translations.json` is written correctly.

Then, we can add a new command to the `template-base`: `npm run parse-translations`, that runs the `parse-translations` script as seen in the deploy command: https://github.com/thoughtindustries/helium/blob/12248ec48a121e5c6b129e3810dd1a7ed9b01a7b/tooling/cli/lib/deploy.js#L155

Then write a similar test case that calls the `npm run parse-translations` and verifies the collected keys from the `template-base` matches what we'd expect.

AlexONeillTI:
I had a sync with Aalto university yesterday and they brought up this translations bug. Is there a possibility of revisiting this to see if it fixes their issue?

@chrisvariety @rjschill87

chrisvariety:
It would be good to have @rjschill87 take another look at it, it seems like there were a lot of commits added after the previous review.

tooling/cli/lib/parse-translations.js
```diff
@@ -58,49 +58,69 @@ const KEYS_WITH_PLURALS = [];
...
-    process.exit(1);
-  });
-
-function addToTranslations(key) {
-  const translatedLanguages = Object.keys(TI_TRANSLATIONS);
-  for (const lang of translatedLanguages) {
-    if (!FINAL_TRANSLATIONS[lang]) {
-      FINAL_TRANSLATIONS[lang] = { lms: {} };
-    }
+  function addToTranslations(key) {
```
rjschill87, line 63: Think it might be easier to follow if you bump these up closer to the top of `processTranslations`, before they're called, since they're scoped to this function.

AlexONeillTI:
I have changes implemented and tested locally. When trying to push, I am getting that I don't have permission to push to helium repo. 

I have messaged @haddadnj and am waiting for a reply. I will update you here when I have successfully pushed the changes.

chrisvariety:
@AlexONeillTI fixed your permissions :)

AlexONeillTI:
> @AlexONeillTI fixed your permissions :)

Thanks Chris!

tooling/cli/lib/deploy.js
```diff
@@ -208,7 +208,7 @@ async function gatherUsedTranslations() {
     const parserScript = path.join(__dirname, 'parse-translations.js');
     const exec = childProcess.exec;
 
-    const parseProcess = exec(`node ${parserScript}`);
+    const parseProcess = exec(`node -e "require('${parserScript}').updateTranslations()"`);
```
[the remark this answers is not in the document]
rjschill87: @AlexONeillTI Apologies but think we got wires crossed. I wasn't necessarily saying to just revert it, just that there was no `updateTranslations` method exported from `parse-translations.js`. Think we would want this to be `mainFunction` now, since that calls the methods previously handled with the self-calling function?

AlexONeillTI:
Review now up to date. Addressed @rjschill87's comment regarding the call to the parserScript here: https://github.com/thoughtindustries/helium/pull/220#discussion_r1757122524

The test suite runs successfully and I also tested the latest changes by deploying an app with the Cart and Dashboard page to my sandbox. Both successfully reflect the translations in locales/translations.json

PR #222, opened 2023-07-28
fix: page path and link refs in starter app

ljiang-ti:
Closes CLM-9205


BReimerTI (approved):

rjschill87:
@ljiang-ti Looks like it wasn't detailed in the ticket, but as we're updating links from `/sign_in` to `/learn/sign_in`, we'll want to move the actual `signin` directory under `/learn` as well I believe.

ljiang-ti:
> @ljiang-ti Looks like it wasn't detailed in the ticket, but as we're updating links from `/sign_in` to `/learn/sign_in`, we'll want to move the actual `signin` directory under `/learn` as well I believe.

thanks @rjschill87 for the notes! Core platform has the `/learn/sign_in` route, by updating the sign in page in helium app, is it intended to override the one in core platform?

ljiang-ti:
@rjschill87 - updated the sign in page path as discussed.

also noted that during local verifications, the learner access widget is not rendering any contents for a learner. It turns out that the component `LoadUserLearning` prop `kind` needs some fixes (might be caused by last refactoring).

rjschill87 (approved):

PR #223, opened 2023-08-10
refactor: server and client side rendering in catalog component

ljiang-ti:
Closes CLM-8911

Couple major changes to the catalog component (~~changes marked with ☑️  are complete~~):
- [X] add (back) support for `layoutId` and `widgetId`
- [X] flatten props for catalog provider
- [X] add prop `ssr` to catalog provider to support server side render and client side render
- [X] update readme
- [X] supplementary updates:
  - [X] update content package for helium query updates (related to https://github.com/thoughtindustries/ti/pull/9723)
  - [X] update packages with graphql codegen setup to account for new TI schema changes (maybe nice-to-haves?)
  - [X] update pagination package to support `ssr` prop
  - [X] update catalog component references due to prop changes (mostly in tooling/template-base)
- [X] additional updates (existing bugs?):
  - [X] fix merged css classes not applying the styles properly (replace `clsx` with `tailwind-merge`) 
  - [X] fix pagination package rendering `Showing ` when server returns zero items

### Server side render behaviors

- Data fetching takes place on server side
- HTML content is rendered on server side
- Anchor links are used for catalog interactions (like changing filters), with its default behavior to do a full page load when clicked
- Minimal scripting on client side
- No concern for loading state

### Client side render behaviors

- Data fetching takes place on client side
- HTML content is rendered on client side
- Anchor links with click handler are used for catalog interactions, upon click, it will trigger client side data fetching
- Heavy scripting on client side
- Need to handle loading state and user experience
- Need to handle browser navigation

## Notes about implementations for `ssr` prop

The catalog package exposes a component `CatalogProvider` to accept the props. Depending on the prop `ssr` value, the component renders 2 internal components. When prop `ssr` is true, the internal component `CatalogServerProvider` is rendered; when prop `ssr` is false, the internal component `CatalogClientProvider` is rendered.

Server side render will behave as-is from component `CatalogServerProvider`.

Client side render is added in component `CatalogClientProvider`. Compared to server side render implementation, this component exposes additional context values for loading state and client side navigation function, as well as handling the browser navigation and DOM element manipulations. Listed below is a breakdown of the implementation:
- flag `ssr` to false in apollo client `useQuery`, to disable server side data fetching.
- track `loading` state
- upon user interaction to change filters, it triggers data fetching and update url with windows History API `pushState`. To smoothen the loading state transitions, scroll browser to the header element and tweak the DOM height (similar to catalog widget in `TI`)
- add event listener to `popstate` to capture browser navigation (like using the `back` or `forward` buttons)

Most of the changes happen in the `CatalogClientProvider` component. To all the UI components, the major change is to swap out the anchor element with a custom `CatalogLinkButton` component. The custom component is a headless component that applies the click event handler depending on the `ssr` context value.


packages/catalog/src/catalog-loader.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactElement } from 'react';
+import { LoadingDots } from '@thoughtindustries/content';
+import { useCatalog } from './core';
+
```
BReimerTI, line 4: @ljiang-ti I think typing this component as follows would be better than as a JSX element since it is taking in props and returning a child component. What do you think?
```suggestion

interface CatalogLoaderProps {
  children: ReactElement;
}

```

packages/catalog/src/catalog-loader.tsx
```diff
@@ -0,0 +1,16 @@
...
+const CatalogLoader = ({ children }: { children: ReactElement }): JSX.Element => {
```
BReimerTI, line 5:
```suggestion
const CatalogLoader = ({ children }: CatalogLoaderProps) => {
```

BReimerTI (approved):

packages/catalog/src/variants/display-type-results/item-queue-button.tsx
```diff
@@ -40,7 +40,7 @@ const ItemQueueButton = ({
     [wasAddedToQueue, isLoading, item, onAddedToQueue]
   );
 
-  const btnClassNames = clsx([
+  const btnClassNames = twMerge([
```
rjschill87, line 43: Think the `twMerge` is here what's causing an issue, more specifically L45. Looks like `twMerge` may not support objects as args this way: https://github.com/dcastil/tailwind-merge/discussions/137#discussioncomment-3481605

PR #224, opened 2023-08-10
fix: helium init fixes

BReimerTI:
A couple of changes related to the authentication and initialization of the Helium app:

- Prevent the `npm run dev` prompt from showing if no instances are authenticated.
- Match community navigation logos with correct urls in the Hello World template.

### DIFF
--- a/tooling/cli/lib/command-modules/authenticate.js
+++ b/tooling/cli/lib/command-modules/authenticate.js
@@ -44,5 +44,9 @@ exports.handler = async function (argv) {
 
     console.log('\nReady to get started? Just run:\n');
     console.log(`$ ${pkgManager} run dev`);
+  } else {
+    console.log(
+      'Error authenticating instance: Please check your instance URL and API key and try again.'
+    );
   }
 };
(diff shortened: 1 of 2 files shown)


tooling/cli/lib/command-modules/authenticate.js
```diff
@@ -39,10 +39,10 @@ exports.handler = async function (argv) {
...
     console.log('Initializing project...');
     await initProject(process.cwd(), instances);
-  }
 
-  const pkgManager = /yarn/.test(process.env.npm_execpath) ? 'yarn' : 'npm';
+    const pkgManager = /yarn/.test(process.env.npm_execpath) ? 'yarn' : 'npm';
 
-  console.log('\nReady to get started? Just run:\n');
-  console.log(`$ ${pkgManager} run dev`);
+    console.log('\nReady to get started? Just run:\n');
```
rjschill87, line 45: If we're moving these in the `if` block, think it'd be helpful to have an `else` to show an error message re: the failure to auth/init instances instead of the command script completing quietly.

rjschill87 (approved):

PR #229, opened 2023-10-20
CLM-9217 Content Header Storybook issue

AlexONeillTI:
Closes #CLM-9217

### DIFF
--- a/packages/content/src/components/content-header/types.ts
+++ b/packages/content/src/components/content-header/types.ts
@@ -1,4 +1,4 @@
-import { ContentKind } from '@thoughtindustries/content/src/graphql/global-types';
+import { ContentKind } from '../../graphql/global-types';
 
 export type ContentHeaderProps = {
   /** Content Kind for content header */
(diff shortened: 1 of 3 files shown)


packages/content/src/components/content-header/types.ts
```diff
@@ -0,0 +1,12 @@
+import { ContentKind } from '@thoughtindustries/content/src/graphql/global-types';
+
+export interface ContentHeaderProps {
```
rjschill87, line 3: Any particular reason to use an `interface` vs `type` here? Also a head's up, since these are the props for the `ContentHeader` component, think it'd be fine to import the type for use here:

https://github.com/thoughtindustries/helium/blob/7ee45955c4672ae73a8d47ba894e7bf546cec999/packages/content/src/components/content-header/content-header.tsx#L43

rjschill87 (commented):
Left a quick question. Also don't think we need to the `package-lock.json` change as there were no dependency changes

PR #230, opened 2023-11-01
Fix: pin vite-plugin-ssr version

rjschill87:
Closes CLM-10059

In version `0.4.142`, `vite-plugin-ssr` was renamed to `vike`. Since our `package.json` ships with the dependency set to `^0.4.43`, more recent installs could end up with` 0.4.142` and – unless some text-replacements are completed (see [[1]](https://vite-plugin-ssr.com/vike) and[[2]](https://github.com/vikejs/vike/discussions/1147#discussioncomment-7112677)) - the project could be compiled without a `/dist/client/manifest.json` and cause deployments to fail.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39801,7 +39801,7 @@
         "remark-mdx-frontmatter": "^2.0.3",
         "tailwindcss": "^3.0.9",
         "vite": "^3.1.8",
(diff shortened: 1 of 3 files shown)


chrisvariety (commented):
thx!

PR #231, opened 2023-11-06
feat: allow deployment of development builds

rjschill87:
Closes CLM-10089

This change allows users to pass a `-d`/`-debug` flag when running `helium deploy`/`npm run deploy` in order to deploy a development build to assist troubleshooting efforts.

In order to do so, users will need to a `.env.development` file with `NODE_ENV='development'`.

Screen Shot 2023-11-06 at 13 32 00

### DIFF
--- a/tooling/cli/lib/file-generators.js
+++ b/tooling/cli/lib/file-generators.js
@@ -5,7 +5,8 @@ const { DEFAULT_GRAPHQL_SOURCE_PATHS } = require('./helpers/constants');
 
 const initProject = async (dir, instances) => {
   console.log('Generating env file...');
-  await generateEnvFile(dir, instances);
+  await generateEnvFile(dir);
+  await generateEnvFile(dir, true);
 
   console.log('Generating translations file...');
   await generateTranslationFile(dir, instances);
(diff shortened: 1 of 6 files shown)


chrisvariety (approved):
🆒 is this everything or is there another PR in another repo?

rjschill87:
@chrisvariety This is the only one. At the moment, I've left the `NODE_ENV` for the CF worker as `'production'`

https://github.com/thoughtindustries/iron-workers/blob/8cbc4a0cee7f1e6a50d691777a2ad8cacf493254/helium-deployer/entrypoint.sh#L78

As that affects whether or not we include the persisted query link w/ GraphQL requests: 

https://github.com/thoughtindustries/helium/blob/888ed632165665c2dc72ba265811c0ca12b83b78/tooling/helium-server/worker/init-page-context.js#L52-L58

tooling/cli/lib/file-generators.js
```diff
@@ -14,13 +15,9 @@ const initProject = async (dir, instances) => {
   return generateConfigFile(dir, instances);
 };
 
-const generateEnvFile = async (dir, instances) => {
-  const fileName = path.resolve(dir, '.env');
-  let data = '';
-
-  instances.forEach(instance => {
-    data += `TI_${instance.nickname}_API_KEY=${instance.apiKey}\n`;
-  });
```
chrisvariety, line 23: Did we intentionally lose this?

rjschill87: Yeah -- it's leftover from when we were going to store all the instance info in a `.env` but opted for `ti-config.json` instead. It's never actually referenced anywhere.

chrisvariety (approved):

PR #233, opened 2023-11-09
fix: temporarily rename env files when using development build commands

rjschill87:
Continues CLM-10089

- Moves `getFileContents` from `/helpers/atoms` to more appropriate `/helpers/filepaths`.
- Updates `deploy` command to temporarily rename `.env.development` to `.env` when deploying development builds, with the files resetting after upload.

### DIFF
--- a/tooling/cli/lib/deploy.js
+++ b/tooling/cli/lib/deploy.js
@@ -174,7 +174,7 @@ async function swapEnvFiles(target, source, newFilePath) {
 
   if (hasTarget && hasSource) {
     const newPath = path.join(OP_DIR, newFilePath);
-    await rename(target, newPath);
+    await rename(targetPath, newPath);
     await rename(sourcePath, targetPath);
   }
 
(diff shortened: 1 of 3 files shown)


chrisvariety (approved):
LGTM, thanks!

tooling/cli/lib/deploy.js
```diff
@@ -140,12 +145,52 @@ process.on('message', message => {
...
+async function swapEnvFiles(target, source, newFilePath) {
+  const targetPath = path.join(OP_DIR, target);
+  const sourcePath = path.join(OP_DIR, source);
+
+  const hasTarget = await getFileContents(targetPath);
+  const hasSource = await getFileContents(sourcePath);
+
+  if (hasTarget && hasSource) {
+    const newPath = path.join(OP_DIR, newFilePath);
+    await rename(target, newPath);
```
chrisvariety, line 177: should this be:

```suggestion
    await rename(targetPath, newPath);
```

?

rjschill87: Funnily enough it worked locally w/ either. Updated for consistency though, wouldn't be surprised to find out it's local configuration/link thing.

PR #234, opened 2023-11-10
fix: catalog page composer

ljiang-ti:
unticketed

Remove the reset params call when composing page urls.


chrisvariety (approved):
thanks much!! 🙏

chrisvariety:
@rjschill87 can you cut a release with this in it? thanks!

PR #237, opened 2024-02-12
feat: add check for HELIUM_PRIVATE_ENABLE_ASSET_PROXY

rjschill87:
Closes CLM-10925

Adds check for `HELIUM_PRIVATE_ENABLE_ASSET_PROXY` env var. 

If `HELIUM_PRIVATE_ENABLE_ASSET_PROXY=true`, we'll leave asset URLs as written and proxy through Thought Industries backend.


chrisvariety (approved):
thx!

PR #239, opened 2024-03-29
Update README.md to include deployment docs

AlexONeillTI:
I recently had two differant customers try to deploy the template-base to their instances. This can lead to some strange behavior and errors. 

The current deployment documentation is here: https://developer.thoughtindustries.com/docs/getting-started and I chatted to @chrisvariety a little while ago about bringing this documentation into the repo also. 

This means that a developer will be able to see how to get a local development server up and running in the template-base, but then will also see they should create their own app and deploy/run a development server for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,54 @@ $ npm run authenticate
 $ npm run dev
 ```
 
+
+## How to deploy a Helium App
+### Step 1: Create a Helium App
+Run the following command in your terminal
+```bash npm2yarn
+ npm init helium-app@latest
+ ```
+This will prompt you to install the `@thoughtindustries/helium` package if you do not already have it installed on your local machine.
+
+### Step 2: Choose a Theme
+Selete a theme to get started building your Helium app.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,6 +42,54 @@ $ npm run authenticate
...
+npm run dev
+ ```
+
+### Step 6: Deploy Your Helium App
+This command will deploy your Helium app to Cloudflare workers managed by Thought Industries. The Helium feature flag must be turned on for you to deploy succesfully.
+```bash npm2yarn
+npm run deploy [nickname]
+```
+
 ### Workspaces
```
chrisvariety, line 93: I think everything below here is related to developing Helium itself, not running Helium. Maybe we can make that clearer in the Readme for folks? Otherwise folks will kind of just go from Step 6 here right into this section.

AlexONeillTI: Good point - Do you think "Helium Development" would work as a header for that section?

PR #240, opened 2024-08-26
TI-626 Routing Issue with Dashboard Bug

madison-kneller-ti:
Resolves [TI-626](https://thoughtindustries.atlassian.net/browse/TI-626)

When replacing the dashboard component, on login, the user would be led to the native dashboard (/learn/) instead of (/learn/dashboard). This fix adds "dashboard" to the end of the `window.location.href` during login to lead the user to the correct dashboard. 

[TI-626]: https://thoughtindustries.atlassian.net/browse/TI-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tooling/template-base/components/Signin/SigninPage.tsx
+++ b/tooling/template-base/components/Signin/SigninPage.tsx
@@ -19,7 +19,7 @@ const SigninPage = () => {
       variables: { email: userEmail, password: userPassword }
     })
       .then(() => {
-        window.location.href = '/learn/';
+        window.location.href = '/learn/dashboard';
       })
       .catch(error => {
         console.log('Handle Login Mutation Error Here');


rjschill87:
@madison-kneller-ti - can you expand a bit on how sending the user to `/learn/dashboard` instead of `/learn` resolves the issue with `/learn` hanging? 

The reason it was set to send users to `/learn` was to replicate native functionality where we'd send students to their dashboard and admins to the manager UI. This change could possibly resolve the issue for users using a custom sign in page - but devs can add a custom dashboard and still use the native sign in, which will still have the "problem" path of sending users to `/learn` after sign in.

madison-kneller-ti:
Closed this PR in favor of a solution in the ti repo

PR #243, opened 2025-03-27
feat: support node 18

ljiang-ti:
Closes TI-1779

We are using Volta to pin the versions of Node and Npm so that they can align with what's used in Helium Deployer. Going to bump the versions to latest Node v18 and respective Npm version. This would allow Helium developers to build the app with Node v18. The minimal Node requirement will remain as-is (v16).


rjschill87:
@ljiang-ti We should probably update the `package.json#engine` properties as well, no?

ljiang-ti:
> @ljiang-ti We should probably update the `package.json#engine` properties as well, no?

@rjschill87 I left the `package.json#engine` as-is, thought updating them to Node v18 will then require the upgrade of Npm packages to Node v18 compatible versions throughout the tool chain in the monorepo (which would be the Option 2 discussed in JIRA). Given the changes here, it will still require minimum Node v16 and is tested with Node 18.

rjschill87 (approved):

PR #247, opened 2025-09-23
chore: update dependencies to resolve security vulnerabilities

ti-jmcevoy:
BREAKING CHANGE: Multiple major version updates that may affect consumers:
- Migrated from vite-plugin-ssr to vike (v0.4.195)
- Updated Storybook from v7 to v8.6.14
- Updated i18next from v21 to v23.16.8
- Updated react-i18next from v11 to v13.5.0

Security fixes (8 vulnerabilities resolved - 3 low, 5 moderate):
- graphql: 16.6.0 → 16.11.0 (uncontrolled resource consumption)
- universal-cookie: ^4.0.4 → ^7.2.2 (cookie vulnerability)
- remark-mdx-frontmatter: ^2.0.3 → ^5.2.0 (prototype pollution)
- tsup: various → ^8.5.0 (esbuild vulnerability)
- Removed unused patch-package (tmp vulnerability)

Package updates across all workspaces:
- vite: ^3.2.8 → ^5.4.11
- @vitejs/plugin-react: ^2.1.0 → ^4.3.4
- lerna: ^4.0.0 → ^8.1.9
- All Storybook packages updated to v8.6.14
- Consistent i18next versions across monorepo

Implementation changes:
- Fixed CJS/ESM compatibility with dynamic Vike imports
- Added deployment manifest generation scripts
- Updated Storybook Apollo addon for v8 compatibility
- Fixed React Hooks violations in i18next decorator
- Added Vite alias for NavBar import resolution
- Updated .gitignore with build artifacts and temp files

Affected packages:
- @thoughtindustries/helium-server
- @thoughtindustries/helium-template
- @thoughtindustries/helium-template-essentials
- @thoughtindustries/storybook-addon-apollo-client
- All @thoughtindustries component packages


tooling/template-base/vite.config.ts
```diff
@@ -1,7 +1,20 @@
...
 import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+import path from 'path';
 
 export default defineConfig(async () => {
+  const plugins = [react()];
+
+  try {
+    // Try to load vike dynamically
+    const vikeModule = await import('vike/plugin');
```
rjschill87: Is there a reason to load vike dynamically here vs importing it from the start? Seems like it could fail silently and SSR would break -- so it'd look like it was working, when it's really limping since it'd be hard to see with it just warning. Feels like something we should surface?

tooling/template-base/vite.config.ts
```diff
@@ -10,7 +23,29 @@ export default defineConfig(async () => {
     rehypePlugins: []
   };
 
-  tiConfig.plugins.push(mdx.default(mdxOptions));
+  plugins.push(mdx.default(mdxOptions));
 
-  return tiConfig;
+  return {
+    plugins,
+    build: {
```
rjschill87: Reason to add these `build` and other props here? Seems like the thing we'd cover in the `vote.config.ts` that ships with `helium-server`? Seems like we could end with issues where someone turns off `manifest`, or changes the `envPrefix` (which I'm pretty sure we wouldn't expect right now) and cause issues?

tooling/helium-server/src/server/server.ts
```diff
@@ -1,12 +1,31 @@
...
 import cookieParser from 'cookie-parser';
-import { renderPage } from 'vite-plugin-ssr/server';
 import findTiInstance from './../utilities/find-ti-instance';
 import { fetchUserAndAppearance, fetchUser } from './../utilities/fetch-user-and-appearance';
 import initPageContext from './../utilities/init-page-context';
 import fetch from 'isomorphic-unfetch';
 import path from 'path';
 
+// Lazy load vike/server to handle CJS/ESM compatibility issues
+let vikeModule: any = null;
```
rjschill87: Think we may want to investigate a different approach here, or at least do some load testing; I'm not sure how Workers handles concurrent requests, but there's potential for a race condition here if it's allowed. Multiple concurrent requests mean multiple imports of `vike/server`.

Is there an issue w/ importing it from the start regardless of env, and instead only overwrite it in dev env?

tooling/template-base/renderer/_default.page.client.tsx
```diff
@@ -66,4 +81,18 @@ async function render(pageContext: PageContext) {
...
   );
+
+  // For Client Routing: check if this is the first render or a navigation
+  if (pageContext.isHydration !== false && !isHydrated) {
+    // Initial page load - hydrate the server-rendered HTML (only once)
+    root = hydrateRoot(pageViewElement, app);
+    isHydrated = true;
+  } else {
+    // Client-side navigation - reuse the root for all subsequent renders
+    if (!root) {
```
rjschill87, line 106: It predates your changes, but think it might be worth implementing the same pattern you've done for `root` for the `apolloClient` -- e.g., creating a `currentApolloClient` so that we can stop/clear the previous Apollo client so we don't get any memory leaks. Something like:

```javascript
if (currentApolloClient) {
  currentApolloClient.stop();
  currentApolloClient.clearStore();
}

currentApolloClient = apolloClient

```

tooling/helium-server/worker/static-assets.js
```diff
@@ -15,8 +19,90 @@ export { handleStaticAssets };
...
+
+      // CRITICAL: Handle Vike's asset URLs that don't have Cloudflare's hash
+      // Vike generates: /assets/entry-server-routing.DTgIhraB.js
+      // Cloudflare has: /assets/entry-server-routing.DTgIhraB.6edf2d3208.js
+      // We need to find the Cloudflare version by matching the Vike pattern
+
+      // Try to match by removing the extension and looking for similar files
+      const pathWithoutExt = cleanPath.replace(/\.(js|css)$/, '');
+
+      for (const [originalPath, hashedPath] of Object.entries(manifest)) {
```
rjschill87, line 68: Did this come from CF or Vice? I'm wondering if there's another way we can implement this -- iterating through the entire manifest on every request seems... expensive. Would it work if we just build a map once when the worker starts? Might want to tighten up the hash matching also, it seems pretty fuzzy.

rjschill87 (commented):
Looks good so far! Still making my way through, but left a few comments on a couple things. Appreciate all the work here!

tooling/template-base/vite.config.ts
ti-jmcevoy: New Implementation:

```node
// In helium-server/src/vite-config/vite.config.ts
export async function createVikeConfig() {
  try {
    const vikeModule = await import('vike/plugin');
    const vike = vikeModule.default || vikeModule;
    
    return mergeConfig(baseConfig, {
      plugins: [vike()]
    });
  } catch (error) {
    console.error('Failed to load vike plugin:', error);
    throw new Error('Vike plugin is required but could not be loaded. Please ensure vike is installed.');
  }
}
```

**Why dynamic import:** We encountered ESM/CJS compatibility issues where Node.js was trying to `require()` an ESM-only module (vike), causing "ERR_REQUIRE_ESM" errors

**Addresses silent failure concern:** We throw an explicit error that stops the build/dev server if vike fails to load

**Clear error messaging:** The error message clearly states what's wrong and what's needed

**No Partial Functionality:** The application won't start without vike, ensuring no partial functionality

These ESM/CJS compatibility issues are really tricky and finicky especially with Vike so this was the best solution I could come up with that didn't break everything.

tooling/template-base/vite.config.ts
ti-jmcevoy: New Implementation:

```node
// In helium-server/src/vite-config/vite.config.ts
export const baseConfig = {
  build: {
    manifest: true,  // Required for production asset mapping
  },
  ssr: {
    noExternal: ['@apollo/client', 'graphql', '@thoughtindustries/helium-*']
  },
  envPrefix: 'VITE_',
  resolve: {
    dedupe: ['graphql', '@apollo/client']
  }
};

// Templates now just extend this:
// In template-base/vite.config.ts
const config = await createVikeConfig();
config.plugins.push(mdx());
```

**Centralized control:** All critical settings are now in `helium-server`, templates can't accidentally break them

**Prevents misconfiguration:** Templates can't turn off `manifest: true` which would break production deployments

**Consistent envPrefix:** Ensures environment variables work consistently across all Helium apps

**Extensible:** Templates can still add their own plugins and settings without overriding critical ones

tooling/helium-server/src/server/server.ts
ti-jmcevoy: New Implementation:

```node
// In helium-server/src/server/server.ts
let renderPage: any = null;  // Module-level cache

export async function setupHeliumServer(...) {
  // Load renderPage dynamically to avoid CJS/ESM issues
  if (!isProduction && viteDevServer) {
    // In development, use dev server's module loader for HMR support
    const vikeDevModule = await viteDevServer.ssrLoadModule('vike/server');
    renderPage = vikeDevModule.renderPage;
  } else if (!renderPage) {  // Only import once in production
    // In production or first time, dynamically import vike
    const vikeModule = await import('vike/server');
    renderPage = vikeModule.renderPage;
  }
}
```

**Race condition prevented:** The `!renderPage` check ensures import happens only once

**Module-level caching:** Once loaded, `renderPage` is cached for all subsequent requests

**Different approach for dev:** Uses Vite's SSR module loader in development for HMR support

**Production optimization:** In production, the import happens once during server startup, not per request

**Worker consideration:** Cloudflare Workers are single-threaded within an isolate, but we still cache to avoid redundant work

tooling/helium-server/worker/static-assets.js
ti-jmcevoy: New Implementation:

```node
// In helium-server/worker/ssr.js
// Parse asset manifest once at worker startup for better performance
// Cloudflare Workers are long-lived, so this cache persists across requests
let CACHED_ASSET_URLS = null;
if (typeof __STATIC_CONTENT_MANIFEST !== 'undefined') {
  try {
    const assetManifest = JSON.parse(__STATIC_CONTENT_MANIFEST);
    CACHED_ASSET_URLS = getAssetUrls(assetManifest);
    console.log('Asset manifest parsed and cached at worker startup');
  } catch (error) {
    console.error('Failed to parse asset manifest at startup:', error);
  }
}

async function handleSsr(url, authToken = null, userAndAppearanceToken = null) {
  // ... 
  const pageContext = await initPageContext(
    url,
    renderPage,
    currentUser,
    appearanceBlock,
    HELIUM_ENDPOINT,
    true,
    sha256,
    authToken,
    null,
    CACHED_ASSET_URLS // Use pre-parsed asset URLs from startup
  );
}
```

**One-time parsing:** Manifest is parsed once when the Worker starts, not on every request

**Long-lived Workers:** Cloudflare Workers persist between requests within an isolate, so the cache remains valid

**Performance improvement:** Eliminates O(n) iteration through manifest on each request

**Memory efficient:** Single shared cache across all requests in the Worker isolate

**Fail-safe:** If parsing fails at startup, it logs the error but doesn't crash the Worker

tooling/template-base/renderer/_default.page.client.tsx
ti-jmcevoy: New Implementation:

```node
// Track current Apollo Client for proper cleanup
let currentApolloClient: ReturnType<typeof makeApolloClient> | null = null;

async function render(pageContext: PageContext) {
  // ... extract pageContext properties

  // Clean up previous Apollo Client to prevent memory leaks
  if (currentApolloClient) {
    // Stop all active queries and subscriptions
    currentApolloClient.stop();
    // Clear the store to free up memory
    await currentApolloClient.clearStore();
    currentApolloClient = null;
  }

  // Create Apollo Client - always create fresh on client side
  const apolloClient = await makeApolloClient(
    heliumEndpoint,
    apolloInitialState,
    isProduction,
    authToken
  );
  
  // Track the current client
  currentApolloClient = apolloClient;
  
  // ... rest of render logic
}
```

**Addresses the Core Concern:** Prevents memory leaks by explicitly cleaning up old Apollo Client instances

**Proper Cleanup Sequence:**
- `stop()` > Terminates all active queries and subscriptions
- `clearStore()` >  Frees cache memory
- `null assignment` > Allows garbage collection

**Applied Consistently:** Implemented in both template-base and template-base-essentials

**Avoids Vike Errors:** Doesn't try to access `pageContext.apolloClient` (which isn't passed to client)

**Performance Impact:** Prevents memory growth in long-running sessions with many navigations

tooling/helium-server/src/vite-config/vite.config.ts
```diff
@@ -1,31 +1,51 @@
...
+// Templates call this to get the complete config
+export async function createVikeConfig() {
   try {
-    // Dynamically import vike to avoid CJS/ESM conflicts
     const vikeModule = await import('vike/plugin');
     const vike = vikeModule.default || vikeModule;
-    baseConfig.plugins.push(vike());
+
+    return {
+      ...baseConfig,
```
rjschill87: May want to double check me but think we'd want to use a deep clone here instead of a spread. Nested objects will still be a reference to the original, so if file A calls `createVikeConfig` and makes a change to `resolve`, then when file B calls `createVikeConfig` later it'll get the version w/ the altered `resolve`.

tooling/helium-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@thoughtindustries/helium-server",
-  "version": "3.0.0-beta.15",
+  "version": "3.0.0-beta.24",
```
rjschill87: More of an all-over thing (or mostly tooling, since components are mostly just react), but w/ all the major changes we might want to update the major version so any folks w/ their dependencies not properly pinned aren't affected.

tooling/helium-server/src/vite-config/vite.config.ts
ti-jmcevoy: Good catch, I've resolved this by using structuredClone() to deep clone all serializable config properties. Here is the new implementation:

```node
// Factory function to create config with vike plugin
// Uses dynamic import to avoid ESM/CJS compatibility issues
// Templates call this to get the complete config
export async function createVikeConfig() {
  try {
    const vikeModule = await import('vike/plugin');
    const vike = vikeModule.default || vikeModule;

    // Deep clone baseConfig to prevent shared references between consumers
    // Note: We can't clone plugins (they contain functions), so we handle them separately
    // This ensures that if one consumer modifies nested objects (e.g., resolve.alias),
    // it won't affect other consumers or the original baseConfig
    const { plugins, ...serializableConfig } = baseConfig;
    const clonedConfig = structuredClone(serializableConfig);

    return {
      ...clonedConfig,
      plugins: [...plugins, vike()] // Use original plugins array (safe - replaced not mutated)
    };
  } catch (error) {
```

tooling/helium-server/package.json
ti-jmcevoy: The current released non-beta version for helium-server is `2.7.0`, at the moment while testing I have been publishing beta packages to NPM so that they can be installed by the helium-deployer for staging instances. I haven't yet published a new production package, but once I do (once this review is complete) I will publish new production packages to NPM with a new major version `3.0.0` due to all of the breaking/large changes and this will be reflected in the package.json file.

rjschill87: @ti-jmcevoy Not quite following, can you clarify? The `helium-deployer` should still be able to install a major change beta fine (for staging at least) without worrying about production at all. Looks like it treats any `Vite` dependency `>=3` as a Worker V2, e.g., it goes off the `Vite` version, not our package version.

ti-jmcevoy: Apologies, I probably wasn't super clear with the above.

Currently in `staging`, `tooling/helium-server/package.json` is pinned to the currently published version:
https://github.com/thoughtindustries/helium/blob/staging/tooling/helium-server/package.json#L3

At the moment this is version `2.7`

When publishing the beta versions to NPM lerna automatically adjusted this for me to the latest beta version that we were currently publishing, which makes sense as the published beta package to NPM should have it's own version properly defined here, I committed this like this because we haven't actually published the final version yet, my plan was to let lerna automatically change this to `3.0.0` once I publish the final package after this review has been approved and then pushing those changes to the branch, but in this case seeing as we know that the final version will be `3.0.0` I can just update this to the final version prior to publishing the package with lerna which should resolve this

ti-jmcevoy: This is now updated with lerna to the correct versions and ready to publish to npm

rjschill87: @ti-jmcevoy -- Looks like doing the major vers change dropped the beta flag? We'll still need that since we'll need to publish packages to test in staging. Been a while but think you might be able to do it with something like 

```sh
lerna version major --conventional-prerelease --message "chore: release new versions" --include-merged-tags
```

But it's been a while so I'd def re-check the docs.

Last question too -- were there any change in the `env` var prefixes that vite/vike uses now? Only asking so they still get picked up by the deployer: 

https://github.com/thoughtindustries/iron-workers/blob/c4f81332c4c17e9a2c26b1e3b1d17425bb321a78/helium-deployer/parse-env-vars.js

rjschill87 (approved):

PR #249, opened 2026-01-14
[TI-3149] - Fix character encoding for special characters

ti-jmcevoy:
Add charset=utf-8 to Content-Type headers for proper special character encoding

Fixes encoding issues where special characters (accented letters, emojis, non-ASCII characters) from external data sources like Contentstack were rendering incorrectly after deployment.

Root cause: HTTP/1.1 defaults to ISO-8859-1 (Latin-1) when charset is not specified in Content-Type header. UTF-8 encoded content was being misinterpreted as Latin-1.

Changes:
- worker/ssr.js: Add charset=utf-8 to text/html and application/json responses
- src/server/server.ts: Add charset=utf-8 to dev server responses for consistency


haddadnj (approved):
LGTM!

